PR #2, opened 2019-06-01
Solidify button appearance combos, document active state

kylesuss:
* Solidify the button `appearance`s -- you can no longer combine multiple string appearances.
* Remove unnecessary `css` calls from within the `styled` blocks
* Pull common string constants out to a variable so that they can easily be changed in the future
* Add PropTypes for the `active` state. I do not really understand the point of this state though and I would like to add a comment for it for docs mode, so I would like to understand the use case
* Add stories for `active` state

--

Over time, I would like to slim up the styles in this component as I think they are a bit hard to follow. It can be hard to tell where certain styles are coming from given there are some duplications in various prop checks and properties. Also, given the styles take up ~300 or so lines, I think there is room to split this out in a way that is a bit easier to track down. It feels like an unnecessary optimization now though, so I believe this would be better served for future work/once it actually becomes a problem.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@storybook/design-system",
-  "version": "0.0.10",
+  "version": "0.0.11",
   "description": "Storybook design system",
   "repository": {
     "type": "git",
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
 import { darken, rgba } from 'polished';
 import { color, typography } from './shared/styles';
 import { easing } from './shared/animation';
@@ -18,13 +18,27 @@ const Loading = styled.span`
   opacity: 0;
 `;
 
+const APPEARANCES = {
+  PRIMARY: 'primary',
+  PRIMARY_OUTLINE: 'primaryOutline',
+  SECONDARY: 'secondary',
+  SECONDARY_OUTLINE: 'secondaryOutline',
+  TERTIARY: 'tertiary',
+  OUTLINE: 'outline',
+};
+
+const SIZES = {
+  SMALL: 'small',
+  MEDIUM: 'medium',
+};
+
 const ButtonWrapper = styled.button`
   border: 0;
   border-radius: 3em;
   cursor: pointer;
   display: inline-block;
   overflow: hidden;
-  padding: ${props => (props.size === 'small' ? '8px 16px' : '13px 20px')};
+  padding: ${props => (props.size === SIZES.SMALL ? '8px 16px' : '13px 20px')};
   position: relative;
   text-align: center;
   text-decoration: none;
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@storybook/design-system",
-  "version": "0.0.10",
+  "version": "0.0.11",
```
kylesuss: I would like to pull this into the learn storybook repo, so I think it needs another release

src/components/Button.js
```diff
@@ -1,6 +1,6 @@
 import React, { Fragment } from 'react';
 import PropTypes from 'prop-types';
-import styled, { css } from 'styled-components';
+import styled from 'styled-components';
```
kylesuss: As mentioned, these `css` calls are not necessary in this context

src/components/Button.js
```diff
@@ -441,16 +464,17 @@ Button.propTypes = {
    Buttons with icons by themselves have a circular shape
   */
   containsIcon: PropTypes.bool,
-  size: PropTypes.oneOf(['small', 'medium']), // this is enum incase we need to add more sizes
+  size: PropTypes.oneOf(Object.values(SIZES)),
```
kylesuss: Thoughts on this approach? The array of values is not immediately apparent, though I like that using the variable means that there is only 1 place where the string value needs to be updated. In docs mode, the array of values looks great.

src/components/Button.js
```diff
@@ -233,6 +251,7 @@ const ButtonWrapper = styled.button`
           &:focus {
             box-shadow: ${color.secondary} 0 0 0 3em inset,
               ${rgba(color.secondary, 0.4)} 0 1px 9px 2px;
+            color: ${color.lightest};
```
kylesuss: To prevent this in the focus state:

Screen Shot 2019-06-01 at 2 59 21 PM

Now it looks like this:
Screen Shot 2019-06-01 at 2 59 10 PM

src/test.js
```diff
@@ -1,7 +1,7 @@
...
+import ExampleComponent from './index';
 
 describe('ExampleComponent', () => {
   it('is truthy', () => {
-    expect(ExampleComponent).toBeTruthy()
-  })
-})
+    expect(ExampleComponent).toBeTruthy();
+  });
+});
```
kylesuss: Prettier updates..

src/components/Button.js
```diff
@@ -421,6 +443,7 @@ export function Button({
 }
 
 Button.propTypes = {
+  active: PropTypes.bool,
```
kylesuss: I tend to prefer using qualifiers for booleans, like `isActive` or `isUnclickable` because I think the cognitive load is lower. It makes it obvious that the value will be true or false without having to look up the docs. That said, I think it should be agreed upon as a standard so I didn't update anything -- just calling it out to get the conversation started.

kylesuss:
@domyen Button updates! Left you a bunch of comments here to get some conversations started. Can you help me out w/ the purpose of the `active` button state as well? I'd like to get some documentation for it in the PropTypes to be clear about the use case.

src/components/Button.js
domyen: That approach sgtm!

src/components/Button.js
domyen: 👍

src/components/Button.js
```diff
@@ -364,64 +297,34 @@ const ButtonWrapper = styled.button`
...
   if (isLink) {
     return (
-      <ButtonLink
-        loading={loading}
-        appearance={appearance}
-        size={size}
-        containsIcon={containsIcon}
-        {...props}
-      >
+      <ButtonLink isLoading={isLoading} disabled={isDisabled} {...props}>
```
kylesuss: `disabled` is always a tricky one w/ regard to the `is` prefixing -- since the default HTML property is just `disabled`, I typically pass that to the actual button element so that it can do things naturally, but require the prefix on the React component that wraps it.

kylesuss:
:rocket: PR was released in v0.0.13 :rocket:

PR #6, opened 2019-06-04
Apply link styles to LinkWrapper

kylesuss:
Note the last 2 (previously the text link had no styling):

Screen Shot 2019-06-03 at 6 33 23 PM

### DIFF
--- a/src/components/StoryLinkWrapper.js
+++ b/src/components/StoryLinkWrapper.js
@@ -1,29 +1,32 @@
 /* eslint-disable import/no-extraneous-dependencies */
 // This is allows us to test whether the link works via the actions addon
-import React, { Children } from 'react';
+import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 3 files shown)


src/components/StoryLinkWrapper.js
```diff
@@ -1,29 +1,32 @@
...
-    href: passHref && href,
-    onClick: e => {
-      e.preventDefault();
-      onLinkClick(href);
-    },
-  });
+  return (
+    <a href={href} {...rest} onClick={modifiedOnClick}>
+      {children}
+    </a>
```
kylesuss: I updated this to better simulate what a `GatsbyLink` (or React Router link for that matter) would return

kylesuss:
@domyen if you merge this, can you let me push the version bump plz? There is a weird thing that I have to do just for this 1 version bump as I made a bit of a mess when I fixed the release process. It won't be an issue after that.

PR #8, opened 2019-06-06
Fix Link transition perf by specifying properties

kylesuss:
`transition: all` can lead to p bad performance and I noticed that some things can be a bit janky as I was importing this into the other app. Better to specify the specific properties you want to transition so the browser doesn't have to guess!

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -8,7 +8,7 @@ import { color } from './shared/styles';
 
 const linkStyles = css`
   display: inline-block;
-  transition: all 150ms ease-out;
+  transition: transform 150ms ease-out, color 150ms ease-out;
   text-decoration: none;
 
   color: ${color.secondary};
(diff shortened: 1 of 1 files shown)


src/components/Link.js
```diff
@@ -8,19 +8,19 @@ import { color } from './shared/styles';
...
   &:hover,
   &:focus {
     cursor: pointer;
-    transform: translate3d(0, -1px, 0);
+    transform: translateY(-1px);
     color: ${darken(0.07, color.secondary)};
   }
   &:active {
-    transform: translate3d(0, 0, 0);
+    transform: translateY(0);
```
kylesuss: Also no need for `translate3d` as the animation looks nice already w/ the updated transition properties specified.

domyen:
Nice!

kylesuss:
:rocket: PR was released in v0.0.19 :rocket:

PR #9, opened 2019-06-06
Export Highlight for consumption

kylesuss:
Missed this last time -- I thought all the components were exported automatically.

### DIFF
--- a/src/components/index.js
+++ b/src/components/index.js
@@ -9,6 +9,7 @@ export * from './Avatar';
 export * from './AvatarList';
 export * from './Badge';
 export * from './Button';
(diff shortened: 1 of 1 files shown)


kylesuss:
:rocket: PR was released in v0.0.20 :rocket:

shilman:
@kylesuss Can we use named exports to keep things consistent? Easier to track through the code 😘

kylesuss:
@shilman Sure! To clarify, are you suggesting that even if there will only ever be 1 export for a file, that we still only export it as a named export?

shilman:
Thanks!! 🙇 That's the one 👍

PR #10, opened 2019-06-06
Fix LinkWrapper props to avoid collisions w React, Gatsby, etc links

kylesuss:
I tried to fix this in an earlier PR, but I didn't get it right. I think I have it now -- A lot of the `Link` components that you use w/ other libs like  Gatsby & React Router complain when you pass them props that are not standard. I kept seeing errors for this in the console:

Screen Shot 2019-06-06 at 5 38 24 PM

I decided to just manually delete the props from the prop object that is passed down to the `LinkWrapper` -- that seemed to do the trick.

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -147,11 +147,8 @@ export function Link({ isButton, withArrow, containsIcon, LinkWrapper, children,
 
   if (LinkWrapper) {
     const StyledLinkWrapper = LinkA.withComponent(LinkWrapper);
-    const linkWrapperProps = { ...rest };
-    delete linkWrapperProps.inverse;
-    delete linkWrapperProps.nochrome;
-    delete linkWrapperProps.secondary;
-    delete linkWrapperProps.tertiary;
+    const { inverse, nochrome, secondary, tertiary, ...linkWrapperProps } = rest;
+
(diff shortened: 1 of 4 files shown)


src/components/Link.js
```diff
@@ -147,12 +147,19 @@ export function Link({ isButton, withArrow, containsIcon, LinkWrapper, children,
 
   if (LinkWrapper) {
     const StyledLinkWrapper = LinkA.withComponent(LinkWrapper);
-    return <StyledLinkWrapper {...props}>{content}</StyledLinkWrapper>;
+    const linkWrapperProps = { ...rest };
+    delete linkWrapperProps.inverse;
+    delete linkWrapperProps.nochrome;
+    delete linkWrapperProps.secondary;
+    delete linkWrapperProps.tertiary;
```
tmeasday: I think a more idiomatic way to do this is:

```js
const { inverse, nochrome, secondary, tertiary, ...linkWrapperProps } = rest;

return <StyledLinkWrapper {...linkWrapperProps}>{content}</StyledLinkWrapper>;
```

kylesuss: Love it. It's funny, I am so used to using that pattern when destructuring in function params, but it felt weird in this context. Makes sense, good eye :)

Updated!

kylesuss:
:rocket: PR was released in v0.0.21 :rocket:

PR #11, opened 2019-06-12
Destructure LinkWrapper props to avoid passing them down

kylesuss:
#10 Didn't work, so I am now trying this. I believe this solves the issue as I added a story for it.


kylesuss:
:rocket: PR was released in v0.0.22 :rocket:

PR #12, opened 2019-06-13
fix: Avatar accessibility

jsomsanith:
Avatar component is not accessible

non-a11y

(SR = Screen Reader)

1. The loading avatar is just skipped by SR (yeah I added a "lol" text on the left to show that the focus switch between the elements between the loading avatar, it's kind of my default text when I don't know what to write :D )
A loader should have aria-busy attribute and a text alternative if it doesn't contains explicit text.

2. The initial avatar alt attribute is not taken into account by SR, it reads only the initial letter

1. Set the aria-busy attribute and a `Loading avatar ...` aria-label

2. Hide the initial with aria-hidden, and set the username in an aria-label instead of alt attribute

a11y

### DIFF
--- a/src/components/Avatar.js
+++ b/src/components/Avatar.js
@@ -103,17 +103,24 @@ const Initial = styled.div`
  */
 export function Avatar({ loading, username, src, size, ...props }) {
   let avatarFigure = <Icon icon="useralt" />;
+  const a11yProps = {};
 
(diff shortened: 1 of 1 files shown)


domyen:
Thank you! This is awesome 🙇

kylesuss:

Issue #13, opened 2019-06-14
Add a11y addon


domyen:
Add the Storybook a11y addon to this project

domyen:
Thanks for enabling the a11y addon @oorestisime!

PR #15, opened 2019-06-15
Enable a11y addon

oorestisime:
Addon was already in the devDependencies so i just enabled it.
Should cover #13


kylesuss:

PR #16, opened 2019-06-15
fix: AvatarList ellipsis in an li

jsomsanith:
#14 fixed the AvatarList semantic with ul/li. But the ellipsis is still in a div.

Use li for the ellipsis instead.


kylesuss:

PR #17, opened 2019-06-15
chore: button icon only demo text alternative

jsomsanith:
Button story has an icon-only button, without text alternative.

Fix the story, adding an aria-label to the icon-only button.


kylesuss:

PR #19, opened 2019-06-17
Add support for closeOnClick to WithTooltip

kylesuss:
Support for `closeOnClick` in `WithTooltp`. Sadly, you have to manage your own state for the tooltip if you want to support this feature. It is not too much additional effort, though, but you have to be careful to put the event handlers in the right places (1 for onVisibilityChange, 1 for the tooltip trigger, and 1 for the tooltip itself).

Gif below showing:
1. Outside click detection still working
2. closeOnClick working when the appropriate prop is passed
3. no closeOnClick behavior when the prop is not passed in

closeOnClick

### DIFF
--- a/src/components/tooltip/WithTooltip.js
+++ b/src/components/tooltip/WithTooltip.js
@@ -1,8 +1,7 @@
-import React from 'react';
+import React, { useState } from 'react';
 import PropTypes from 'prop-types';
 import styled from 'styled-components';
 import TooltipTrigger from 'react-popper-tooltip';
-import { withState } from 'recompose';
 
 import { Tooltip } from './Tooltip';
 
@@ -25,17 +24,27 @@ function WithTooltip({
   hasChrome,
   tooltip,
   children,
-  tooltipShown,
+  startOpen,
   onVisibilityChange,
   ...props
 }) {
   const Container = svg ? TargetSvgContainer : TargetContainer;
+  const [isTooltipShown, setTooltipShown] = useState(startOpen);
+  const toggleTooltipShown = () => setTooltipShown(!isTooltipShown);
+  const closeTooltipOnClick = () => {
+    if (!closeOnClick) {
+      return;
+    }
+
+    setTooltipShown(false);
+  };
+
   return (
     <TooltipTrigger
       placement={placement}
       trigger={trigger}
-      tooltipShown={tooltipShown}
-      onVisibilityChange={onVisibilityChange}
+      tooltipShown={isTooltipShown}
+      onVisibilityChange={toggleTooltipShown}
       modifiers={modifiers}
       tooltip={({
         getTooltipProps,
@@ -50,6 +59,7 @@ function WithTooltip({
           tooltipRef={tooltipRef}
           arrowRef={arrowRef}
           arrowProps={getArrowProps()}
+          onClick={closeTooltipOnClick}
           {...getTooltipProps()}
         >
           {typeof tooltip === 'function'
@@ -59,7 +69,7 @@ function WithTooltip({
       )}
     >
       {({ getTriggerProps, triggerRef }) => (
-        <Container ref={triggerRef} {...getTriggerProps()} {...props}>
+        <Container ref={triggerRef} {...getTriggerProps()} {...props} onClick={toggleTooltipShown}>
           {children}
         </Container>
       )}
@@ -76,7 +86,7 @@ WithTooltip.propTypes = {
   hasChrome: PropTypes.bool,
   tooltip: PropTypes.oneOfType([PropTypes.node, PropTypes.func]).isRequired,
   children: PropTypes.node.isRequired,
-  tooltipShown: PropTypes.bool,
+  startOpen: PropTypes.bool,
   onVisibilityChange: PropTypes.func.isRequired,
 };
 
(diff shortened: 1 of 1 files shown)


kylesuss:
@tmeasday word on the street is that you are the resident tooltip expert 😉 

Any thoughts on how to effectively capture clicks on links in the tooltip in order to close the tooltip when they are clicked? I understand this feature used to work before the design system was implemented, though it seems the libraries in use are slightly different (`react-popper-tooltip` now vs `react-popper` previously).

Right now this PR implements a dumb solution that captures a click to _anything_ in the tooltip, which is not ideal. I have a few other ideas to capture clicks to links only, but neither of them seem ideal either:

**1. Setup a handler of some sort on our `TooltipLinkList` or similar components that can be used to manually call the function to close the tooltip onClick**

This has the downside that you have to use a link from the design system, which I don't think is always going to be realistic.

**2. Setup some sort of logic in the `WithTooltip` function that manually adds an event handler for links and buttons that are children of the current dom node. This feels very "un-react-y" so I hesitate to do this, but it might be the most accurate option that I can think of.**

Whatcha think?

tmeasday:
So the older implementation was a bit "dumber" in that it treated *any* click inside the tooltip as a trigger to close it. I can see the argument that that isn't the greatest idea. Our logic was pretty simple, and boiled down to a click hander on the body with the following basic logic:

```js
    if (closeOnClick) {
      this.onHide();
    } else {
      // call onHide() if the click was outside the tooltip.
      this.onHideIfOutsideTooltip(event);
    }
```

The above could probably be improved by simply looking at the `event.target` and checking if it is a `<a>` or `<button>` or a descendant of one, which would achieve the effect you are looking for. Do we already set a global event handler (e.g. on the `body`?) -- how do we implement the "close on click outside"?

I'm pretty unsure about the approach you have taken here -- the issue is what happens if/when the tooltip contents re-render -- your click handlers will be broken or you may miss legitimate targets. Or does `useEffect` handle this use case assuming any re-rendering is handled by React?

I do kind of feel like the jquery-style "bind at the top level, check `event.target`" approach will work and is in many ways simpler.

kylesuss:
@tmeasday 

> Do we already set a global event handler (e.g. on the body?) -- how do we implement the "close on click outside"?

`react-popper-tooltip` does this for us actually -- for the most part. Given we have to manage the tooltip state ourselves now, we have to [manually set the state back to false](https://github.com/storybookjs/design-system/blob/cccadb01a8bf72ee940439aeed3cf00be9c71779/src/components/tooltip/WithTooltip.js#L54) but the lib provides the `onVisibilityChange` handler which is called in this circumstance.

I'd agree that the logic to capture _any_ click in the tooltip is way easier to manage. That is how I originally did it. @domyen  and I were chatting about the experience though and wanted to try to see if we could get something working that was more aware of what was clicked.

Re: your example:

> The above could probably be improved by simply looking at the event.target and checking if it is a <a> or <button> or a descendant of one, which would achieve the effect you are looking for.

I've done things like this in the past and things have gotten messy when there are a few nested `span`s or something underneath the clickable element (I believe we would run into this as well). How many levels up do you need to go to determine "is this thing a link?" -- it can feel arbitrary though I was just checking the `parentElement` a few times. Maybe you know of a better way?

Re: the `useEffect` approach, [you specify the state/props values that matter to the effect](https://github.com/storybookjs/design-system/blob/cccadb01a8bf72ee940439aeed3cf00be9c71779/src/components/tooltip/WithTooltip.js#L47) in order to let React know when/if it should re-run and the [return value of the function](https://github.com/storybookjs/design-system/blob/cccadb01a8bf72ee940439aeed3cf00be9c71779/src/components/tooltip/WithTooltip.js#L45) is actually the cleanup function to remove the old event listeners (some docs on that here: https://reactjs.org/docs/hooks-effect.html#example-using-hooks-1). So this solution is not going to be able to handle dynamic updates to the tooltip (like adding a link after the tooltip opens) but it does handle just opening/closing a tooltip with links in it just fine. Note sure how common of a use case it would be to have dynamic changes inside of a tooltip -- do you think this is a concern?

Again, not tied to any solution, just didn't want to have to write a thing that says "check up the chain 4 levels to see if this thing is a descendant of a link" as I don't know of a good way to determine that value (4 in the example) so let me know if you have some tricks to make that easy.

tmeasday:
> I've done things like this in the past and things have gotten messy when there are a few nested spans or something underneath the clickable element (I believe we would run into this as well). How many levels up do you need to go to determine "is this thing a link?" -- it can feel arbitrary though I was just checking the parentElement a few times. Maybe you know of a better way?

In the past I just kept checking all the way up the chain of parents until I ran out (i.e. reached the body). Do you see an issue with this?

> Not sure how common of a use case it would be to have dynamic changes inside of a tooltip -- do you think this is a concern?

I'm pretty sure I have run into this issue before -- the case was a tooltip that had loading state inside, then later rendered content. In Chromatic we have quite a few complex tooltips that are also stateful and rerender -- for instance the branch/build picker on [this screen](https://www.chromaticqa.com/library?appId=5a375b97f4b14f0020b0cda3):

image

kylesuss:
That's good feedback re: those complex tooltips. I agree that would be a challenge to handle with the `useEffect` approach.

Re: just going up the chain until you hit the `body`, I am coming around to the idea for sure. Mentally it sounds slightly inefficient given the element could be nested pretty deep but as long as it's not a perf concern, I think it makes the most sense now as well. I'll give it a go tomorrow and see if it feels sluggish for any reason.

Thanks for the feedback!

tmeasday:
> as long as it's not a perf concern

Nah, if it is responding to user input (i.e. a click) and not happening all the time (e.g. scrolling or something) then you can afford to do a bit of work

kylesuss:
@tmeasday @domyen I switched the approach around -- feeling good!

clicky

Not that much work, Tom. Nice call.

This is ready for review again!

domyen:
boo ya! thanks @kylesuss

kylesuss:

PR #20, opened 2019-06-18
fix: invalid input with aria-invalid attribute

jsomsanith:
Input in error should have the `aria-invalid` attribute, so assistive technologies announce it as invalid user entry.

I forgot that in PR #18, sorry for that

Add the `aria-invalid` attribute on the input when we have an associated error.


domyen:
Nice, lgtm

kylesuss:

PR #21, opened 2019-06-18
Improve ListItem

domyen:
- Ensure that it doesn't look broken when there are long text strings
- Fix loading state
- Stories more representative of actual usage


kylesuss:

PR #22, opened 2019-06-18
fix: Textarea accessibility

jsomsanith:
Textarea component is not accessible.
1. label semantic is not correct.
2. labels are not associated to the textarea. Focusing on the textarea, SR won't read the label.
3. labels are not required.
4. invalid textarea should have aria-invalid attribute.
5. errors are not associated to textarea. Focusing on the textarea, SR won't read the error.
6. error is before the textarea. SR will read it, but blind people don't know that there is an associated textarea until they focus on the next element. It lacks some context.
7. subtext is not associated to the textarea

1. use `<label>`
2. accept a required props `id` that is passed to the textarea. Use `htmlFor` react attribute to associate the label with the textarea through the id.
3. labels are required, introducing a `hideLabel` props if we want it not to be displayed.
4. add aria-invalid attribute on textarea with error
5. generate an error id based on the props.id, that is passed to the error element. Use `aria-describedby` on the textarea to associate the error with it.
6. hide the error from SR. It is still associated to the textarea, focusing on it will make SR read the error too. 
7. same as error, generate an id and associate it via `aria-describedby`

### DIFF
--- a/src/components/Textarea.js
+++ b/src/components/Textarea.js
@@ -23,9 +23,6 @@ const Label = styled.label`
 const ErrorMessage = styled.span`
   color: ${color.negative};
   font-weight: ${typography.weight.regular};
-  &:not(:only-child) {
-    margin-left: 0.5em;
-  }
 `;
 
 const LabelWrapper = styled.div`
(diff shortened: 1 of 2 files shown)


domyen:
Nice! I'm going to merge this. I think the inputs and textareas are complex enough to warrant documenting them via Storybook Docs. That would help articulate the finer points of usage and accessibility attrs `aria-...`.

kylesuss:

PR #23, opened 2019-06-20
fix: checkbox accessibility

jsomsanith:
Checkbox component is not accessible
1. non focusable and ignored by SR because of `visibility: hidden`
2. label is not required
3. errors are set as labels
4. checkbox with errors are not set as invalid
5. missing focus visual aspect for keyboard navigation

1. make input visible again, but outside of viewport
2. label is required, introducing `hideLabel` props. Be careful to give a clear context when you hide the label in the designs
3. errors are in a div next to the label, associated with input via `aria-describedby`
4. use `aria-invalid` on input when there is an error
5. TODO

### DIFF
--- a/src/components/Checkbox.js
+++ b/src/components/Checkbox.js
@@ -79,6 +79,10 @@ const Input = styled.input.attrs({ type: 'checkbox' })`
     box-shadow: ${color.mediumdark} 0 0 0 1px inset;
   }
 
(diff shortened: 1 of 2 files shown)


domyen:
Thanks, I'll work on 5 tomorrow and commit updates to this branch 👍

kylesuss:
:rocket: PR was released in v0.0.26 :rocket:

PR #24, opened 2019-06-20
Only pass appropriate link props to Tooltip ListItem

kylesuss:
This is to avoid PropType errors like this:

Screen Shot 2019-06-20 at 2 33 40 PM

Links are weird about passing things down.


kylesuss:
:rocket: PR was released in v0.0.25 :rocket:

PR #25, opened 2019-06-23
fix: radio accessiblity

jsomsanith-tlnd:
Radio component is not accessible
1. non focusable and ignored by SR because of `visibility: hidden`
2. label is not required
3. errors are set as labels
4. checkbox with errors are not set as invalid
5. missing focus visual aspect for keyboard navigation
6. subLabels are set as labels

1. make input visible again, but outside of viewport
2. label is required, introducing `hideLabel` props. Be careful to give a clear context when you hide the label in the designs
3. errors are in a div next to the label, associated with input via `aria-describedby`
4. use `aria-invalid` on input when there is an error
5. use the same style as checkbox for focus state
6. subLabels are not linked as descriptions. Should we rename it ?

### DIFF
--- a/src/components/Radio.js
+++ b/src/components/Radio.js
@@ -1,25 +1,56 @@
 import React from 'react';
 import PropTypes from 'prop-types';
-import styled from 'styled-components';
+import styled, { css } from 'styled-components';
+import { rgba } from 'polished';
 import { color, typography } from './shared/styles';
 
 const Label = styled.label`
   cursor: pointer;
   font-size: ${typography.size.s2}px;
   font-weight: ${typography.weight.bold};
-  min-height: 1em;
   position: relative;
-  display: block;
+  display: inline-block;
+  height: 1em;
+`;
+
+const OptionalText = styled.span`
+  display: inline-block;
+  line-height: 1em;
+  ${props =>
+    props.hideLabel &&
+    css`
+      border: 0px !important;
+      clip: rect(0 0 0 0) !important;
+      -webkit-clip-path: inset(100%) !important;
+      clip-path: inset(100%) !important;
+      height: 1px !important;
+      overflow: hidden !important;
+      padding: 0px !important;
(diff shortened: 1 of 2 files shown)


kylesuss:
🎉 

Thanks for another contribution @jsomsanith-tlnd ! This is great.

I see a slight difference in the alignment of the description (the old subLabel):

https://www.chromaticqa.com/snapshot?appId=5ccbc373887ca40020446347&id=5d0fee6c77b75e0020f10492

Looks left aligned now vs offset from the left. Is that something you can take a look at before we merge this?

src/components/Radio.js
```diff
@@ -1,49 +1,75 @@
...
+      border: 0px !important;
+      clip: rect(0 0 0 0) !important;
+      -webkit-clip-path: inset(100%) !important;
+      clip-path: inset(100%) !important;
+      height: 1px !important;
+      overflow: hidden !important;
+      padding: 0px !important;
+      position: absolute !important;
+      white-space: nowrap !important;
+      width: 1px !important;
```
kylesuss: Were these `!important` tags necessary to make the styles apply correctly? Or are they more for being explicit about the behavior? Does look to me like they should work without `!important`, but maybe I am missing something.

domyen: This is extra safety to make sure no inherited styles collide. 👍

kylesuss:
:rocket: PR was released in v0.0.26 :rocket:

src/components/Radio.js
jsomsanith: @kylesuss Just for extra info, this part has `!important` which is usually veeeeery bad, but this is a standard trick to make it rendered out of the viewport, but still visible by screen readers.
Every css framework has a class with this type of css, for example bootstrap has `sr-only`.

kylesuss: Nice. Thanks for the explanation @jsomsanith . Makes sense!

PR #27, opened 2019-06-24
fix: select accessibility

jsomsanith:
Select component is not accessible
1. label is not required
2. errors are not associated to the select
3. selects with errors are not set as invalid
4. icon is decorative but not hidden to SR
5. progress status is only visual

1. label is required, and associated to the select via `for` attribute. We introduce the `hideLabel` props. Be careful to give a clear context when you hide the label in the designs
2. errors are in a div next to the label, associated with select via `aria-describedby`
3. use `aria-invalid` on select when there is an error
4. set `aria-hidden` to the icon
6. progress has a `Loading` label, that is associated with the select as a descriptionn, so screen reader will read it on focus. The select has an `aria-busy` attribute too

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -6,13 +6,28 @@ import { jiggle } from './shared/animation';
 import { Icon } from './Icon';
 import { Spinner } from './Spinner';
 
-const Label = styled.span`
+const Label = styled.label`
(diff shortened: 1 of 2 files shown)


kylesuss:

PR #28, opened 2019-06-24
fix: icon story accessibility

jsomsanith:
Icon story is not accessible. This is only on the story as we can’t do anything to ensure icon accessibility, it’s more linked to the usage.
On decorative icons we should hide it via `aria-hidden`, and on icons that hold information, we should set a proper `aria-label`.

1. it’s a list but it only uses divs
2. story with label has `decorative` icons, they should be hidden to SR
3. story without labels icons hold information, they should have some text alternatives
4. stories with inline/block icon should have a text alternative

1. use ul/li
2. add aria-hidden on the icons
3. set key as `aria-label`
4. set a meaningful `aria-label`

As suggested by @domyen, a comment in Icon code has been added, giving advice to make it accessible when using it.

### DIFF
--- a/src/components/Icon.js
+++ b/src/components/Icon.js
@@ -15,6 +15,12 @@ const Path = styled.path`
   fill: currentColor;
 `;
 
+/**
+ * An Icon is a piece of visual element, but we must ensure its accessibility while using it.
+ * It can have 2 purposes:
+ * - decorative only: for example, it illustrate a label next to it. We must ensure that it is ignored by screen readers, by setting `aria-hidden` attribute (ex: <Icon icon="check" aria-hidden />)
+ * - non-decorative: it means that it delivers an information. For example, an icon as only child inn a button. The meaning can be obvious visually, but it must have a proper text alternative via `aria-label` for screen readers. (ex: <Icon icon="print" aria-label="Print this document" />)
+ */
 export function Icon({ icon, block, ...props }) {
   return (
     <Svg viewBox="0 0 1024 1024" width="20px" height="20px" block={block} {...props}>
--- a/src/components/Icon.stories.js
+++ b/src/components/Icon.stories.js
@@ -6,11 +6,11 @@ import { Icon } from './Icon';
 import { icons } from './shared/icons';
 
 const Meta = styled.div`
-  color: #999;
+  color: #666;
   font-size: 12px;
 `;
 
-const Item = styled.div`
+const Item = styled.li`
   display: inline-flex;
   flex-direction: row;
   align-items: center;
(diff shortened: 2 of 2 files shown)


kylesuss:
I know its a bit late now as this is merged, but I was sketching out an idea to see if we could make this more enforceable based upon your comments @jsomsanith :

```javascript
const NonDecorativeIcon = ({ label, ...rest }) => (
  <Svg viewBox="0 0 1024 1024" width="20px" height="20px" aria-label={label} {...props}>
    <Path d={icons[icon]} />
  </Svg>
);

NonDecorativeIcon.propTypes = {
  label: PropTypes.string.isRequired,
};

const DecorativeIcon = props => (
  <Svg viewBox="0 0 1024 1024" width="20px" height="20px" aria-hidden {...props}>
    <Path d={icons[icon]} />
  </Svg>
);

const Icon = ({ hasExternalLabel, ...rest }) => {
  if (hasExternalLabel) {
    return <NonDecorativeIcon {...rest} />;
  }

  return <DecorativeIcon {...rest} />;
};

Icon.propTypes = {
  // This is required so that we don't assume all icons are decorative.
  hasExternalLabel: PropTypes.bool.isRequired,
  icon: PropTypes.string.isRequired,
};

export default Icon;
```

@jsomsanith what do you think about something like this? We can enforce the accessibility attributes this way, but is it overkill?

cc @domyen

domyen:
I think it replaces one set of attributes to remember with props. Even
though it makes it somewhat easier to enforce I feel like going with the
standard aria is simpler to remember.

jsomsanith:
Instead of having an Icon element that switch to Decorative or NonDecorative, I would prefer to expose only the `DecorativeIcon` and `NonDecorativeIcon`. So the dev must pick one depending on the purpose. Those 2 components would use the same `Icon` internal component passing the extra `aria-label` or `aria-hidden` from their own required props

kylesuss:
I think the aria stuff is hard to remember if you don't even know to consider using it in the first place. Pretty easy to just leave it out at that point. That said, I don't love the verbose nature of requiring the `hasExternalLabel` prop either -- was just trying to think this through a bit in terms of how the library will be used. It does seem too easy to me to leave the accessibility stuff out. I also prefer having 1 `Icon` component (vs exposing both `DecorativeIcon` and `NonDecorativeIcon`), so maybe that means we have the easiest solution so far and we just have to rely on people checking the design system story docs before using a component.

Seems like a prime use case for docs mode w/ MDX 😉  I think I will use this as a testing ground for some of that documentation.

kylesuss:

PR #29, opened 2019-06-25
Refactor Highlight CSS loading to avoid css-loader expectation

kylesuss:
An interesting thing happened when I was trying to import this design system into another app -- the CSS file in the `Highlight` component was throwing errors as if it could not be loaded properly. I've realized that there is not an issue w/ this in Storybook as they have their own CSS loader and so does Gatsby (which is why it didn't come up during the learnstorybook.com development).

Therefore, I have decided to just copy over the prism CSS theme rather than try to import the CSS file for a few reasons:

1. The build process in this repo does not utilize webpack, so it would be more difficult to add that in addition to the `css-loader`
2. There did not seem to be a babel equivalent to webpack's `css-loader` that I could use instead (we build the design system w/ babel and not webpack)
3. It seems like we should not assume that the consumers of this design system should have to set up the `css-loader`

### DIFF
--- a/src/components/Highlight.js
+++ b/src/components/Highlight.js
@@ -5,15 +5,19 @@ import styled from 'styled-components';
 import Prism from 'prismjs';
 import loadLanguages from 'prismjs/components/index';
 import PropTypes from 'prop-types';
-
 import { color, typography } from './shared/styles';
-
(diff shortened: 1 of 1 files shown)


kylesuss:

PR #33, opened 2019-06-26
Fix Input, TextArea label font weight

kylesuss:
`extrabold` does not exist anymore, so the `Input` and `TextArea` were not using the appropriate var for setting their font weight.

Found a few other places w/ the same issue.


kylesuss:
:rocket: PR was released in v0.0.29 :rocket:

PR #34, opened 2019-06-26
fix: link stories accessibility

jsomsanith:
Link stories are not accessible. This is only on the stories.
1. links with icon only don’t have text alternative
2. color contrast is not correct
3. last example (ListWrapper) is weird, having a button inside an anchor

1. add aria-label on link and add documentation
2. TODO
3. TODO: change the story

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -135,6 +135,9 @@ const LinkButton = styled.button`
   ${linkStyles};
 `;
 
+/**
+ * Links can contains text and/or icons. Be careful using only icons, you must provide a text alternative via aria-label for accessibility.
+ */
 export function Link({ isButton, withArrow, containsIcon, LinkWrapper, children, ...rest }) {
   const content = (
     <Fragment>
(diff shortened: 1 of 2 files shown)


jsomsanith:
@domyen could you update the colors to have a better contract, especially on button active state (blue text on blue background)

kylesuss:
@jsomsanith Just made an issue for this comment:

> last example (ListWrapper) is weird, having a button inside an anchor

https://github.com/storybookjs/design-system/issues/36

It won't be as simple as just fixing the story as we need that functionality, but as you noticed, it should be done differently.

domyen:
@jsomsanith and I chatted about the `LinkWrapper` implementation (used for Gatsby/Next/React router). We agreed to punt on it for now because it's an implementation detail that gets resolved once you put this component in a framework.

@kylesuss wrote up #36 to include `LinkWrapper` prop natively in the `Button` component so we don't need this hack. 

Merging this work for now and will revisit later as part of the ticket.

kylesuss:

PR #35, opened 2019-06-27
Add lightningoff icon

kylesuss:
`lightningoff` was not apart of the design system but is used in chromatic, so I feel like it makes sense to put it in here since `lightning` exists too.


kylesuss:
:rocket: PR was released in v0.0.30 :rocket:

PR #37, opened 2019-06-27
Fix tertiary hover styles. Use tertiary color

kylesuss:
Fixing a small issue w/ the tertiary button hover style. It was using the secondary color rather than the tertiary color.


domyen:
Nice

kylesuss:

PR #38, opened 2019-06-27
Fix Link containsIcon specificity

kylesuss:
The way that the `containsIcon` styles were being applied was having an effect on their priority and therefore, they were never being utilized (they are the crossed out ones):
Screen Shot 2019-06-27 at 11 14 45 AM

I moved them around a bit so that they will take priority when that prop is passed in.


kylesuss:

PR #39, opened 2019-06-27
Export spinner so it can be consumed

kylesuss:
Spinner was not being exported, so it threw some errors when I tried to import and use it.


kylesuss:
:rocket: PR was released in v0.0.32 :rocket:

PR #40, opened 2019-06-27
Export tooltip ListItem for consumption

kylesuss:
Going to need the `ListItem` exported in order to use it 😄


kylesuss:
:rocket: PR was released in v0.0.32 :rocket:

Issue #36, opened 2019-06-27
Support LinkWrapper in Button component


kylesuss:
Similar to how we support the concept of a `LinkWrapper` in our [`Link` component](https://github.com/storybookjs/design-system/blob/1e791e3cab695855f8d6748bd5a596ffbeaab326/src/components/Link.js#L148), we need the same concept in the `Button` component.

Sometimes something should visually look like a button but also be a link. We support that right now with the `isLink` prop on `Button`, however it does not cover the case where you need to use a special Link component (from Gatsby, Next, React Router, etc.).

We have a less than ideal workaround for this now, which is to put a `Button` inside of a `Link`, but the `LinkWrapper` solution discussed here is more ideal and accessible per #34 .

jsomsanith:
Done with #41

PR #41, opened 2019-06-28
feat: ButtonWrapper prop in Button to customise rendering

jsomsanith:
This PR resolve PR #36.
The Link component has a `LinkWrapper` feature, that takes a component to render, adding a Link style on it.
We want the same feature for button.

`Button` component now has a `ButtonWrapper` props, allowing to write that :
```
<Button ButtonWrapper={CustomButton} appearance="primary">
        Primary
</Button>

<Button ButtonWrapper={CustomAnchor} appearance="primary" href="http://storybook.js.org">
        Primary
</Button>
```

This will render `CustomButton`/`CustomAnchor` applying the `Button` style

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -32,7 +32,7 @@ const SIZES = {
   MEDIUM: 'medium',
 };
 
-const ButtonWrapper = styled.button`
+const StyledButton = styled.button`
   border: 0;
   border-radius: 3em;
   cursor: pointer;
@@ -305,9 +305,17 @@ const ButtonWrapper = styled.button`
 
 `;
 
-const ButtonLink = ButtonWrapper.withComponent('a');
+const ButtonLink = StyledButton.withComponent('a');
 
-export function Button({ isDisabled, isLoading, loadingText, isLink, children, ...props }) {
+export function Button({
+  isDisabled,
+  isLoading,
+  loadingText,
+  isLink,
+  children,
+  ButtonWrapper,
+  ...props
+}) {
   const buttonInner = (
     <Fragment>
       <Text>{children}</Text>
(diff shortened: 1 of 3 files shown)


domyen:
Awesome! Can you update the Button story so folks can find out about the usage?

jsomsanith:
Oh, I added 2 stories, guess I haven’t pushed them

jsomsanith:
That’s it :D I just pushed

jsomsanith:
@kylesuss fixed, and I added an optimisation: we memoize the styled button wrapper, instead of creating a new component at each render. Everytime it’s a new component, React considers it as a component change, the old will always be unmounted, and a new one mounted instead.

src/components/Button.js
```diff
@@ -305,27 +305,46 @@ const ButtonWrapper = styled.button`
...
 `;
 
-const ButtonLink = ButtonWrapper.withComponent('a');
+const ButtonLink = StyledButton.withComponent('a');
 
-export function Button({ isDisabled, isLoading, loadingText, isLink, children, ...props }) {
+const applyStyle = ButtonWrapper => {
+  return (
+    ButtonWrapper &&
+    StyledButton.withComponent(({ containsIcon, isLoading, isUnclickable, ...rest }) => (
```
kylesuss: Ah cool! Didn't realize you could do this same destructuring technique using `withComponent`.

src/components/Button.js
```diff
@@ -305,27 +305,46 @@ const ButtonWrapper = styled.button`
...
     </Fragment>
   );
 
-  if (isLink) {
-    return (
-      <ButtonLink isLoading={isLoading} disabled={isDisabled} {...props}>
-        {buttonInner}
-      </ButtonLink>
-    );
+  const StyledButtonWrapper = React.useMemo(() => applyStyle(ButtonWrapper), [ButtonWrapper]);
```
kylesuss: Nice idea w/ the memoization 👌

kylesuss:
:rocket: PR was released in v0.0.35 :rocket:

PR #42, opened 2019-06-28
Refactor tooltip ListItem for flexibility in to/href prop names

kylesuss:
The `tooltip/ListItem` and `tooltip/TooltipLinkList` currently make a lot of assumptions about the prop name that contains the "href" for a link. This is fine for standard links which always use an href, but it complicates things when you also pass in a `LinkWrapper`. Gatsby links use the `to` prop whereas Next links use the `href` prop, so the code that we had:

```
href={!LinkWrapper ? href : undefined}
to={LinkWrapper ? href : undefined}
```

... is making too many assumptions about what every single `LinkWrapper` expects. I propose we leave those assumptions out and instead just pass the props down to the `LinkWrapper` so that this component can be more flexible. This means that whenever you use the `ListItem`, you have to be careful about what props you pass down, however that's the same concern you will have when you are mixing Next links and anchor tags anyway.

### DIFF
--- a/src/components/tooltip/ListItem.js
+++ b/src/components/tooltip/ListItem.js
@@ -52,8 +52,7 @@ const ItemInner = styled.span`
   }
 `;
 
-// eslint-disable-next-line jsx-a11y/anchor-has-content
-const Item = styled(({ active, loading, ...rest }) => <a {...rest} />)`
+const linkStyles = css`
   font-size: ${typography.size.s1}px;
   transition: all 150ms ease-out;
   color: ${color.mediumdark};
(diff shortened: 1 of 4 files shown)


src/components/tooltip/ListItem.js
```diff
@@ -151,19 +143,17 @@ export function ListItem({
...
+    const StyledLinkWrapper = styled(({ active, loading, ...linkWrapperRest }) => (
+      <LinkWrapper {...linkWrapperRest} />
+    ))`
+      ${linkStyles};
+    `;
+
+    return <StyledLinkWrapper {...rest}>{linkInner}</StyledLinkWrapper>;
+  }
+
+  return <Item {...rest}>{linkInner}</Item>;
```
kylesuss: This has come up a few times, but this extra work is needed to avoid passing down non-standard props to `<a>` tags.

src/components/tooltip/ListItem.js
```diff
@@ -151,19 +143,17 @@ export function ListItem({
...
-      active={active}
-      loading={loading}
-      disabled={disabled}
-      {...props}
-    >
-      {linkInner}
-    </Item>
-  );
+  if (LinkWrapper) {
+    const StyledLinkWrapper = styled(({ active, loading, ...linkWrapperRest }) => (
```
kylesuss: This has come up a few times, but this extra work is needed to avoid passing down non-standard props to `<a>` tags.

kylesuss:
:rocket: PR was released in v0.0.33 :rocket:

PR #44, opened 2019-06-28
fix: Link - LinkWrapper optimisation

jsomsanith:
Link component accepts a `LinkWrapper` props. It is styled via styled-component, but it’s not exactly `LinkWrapper` that is passed, but a new component that omits a set of props.
But everytime `Link` renders, the styled `LinkWrapper` is created again, defining a new component. 

React considers it as a new component, the old button will be unmounted, and the new one will be mounted, at every render.

Memoize the styled `LinkWrapper` component.

Bonus: the weird `Link` story with a button as child is removed, inn favor of `Button` component with `ButtonWrapper`

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -144,6 +144,17 @@ const LinkButton = styled.button`
   ${linkStyles};
 `;
 
+const applyStyle = LinkWrapper => {
+  return (
+    LinkWrapper &&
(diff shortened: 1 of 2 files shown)


kylesuss:
:rocket: PR was released in v0.0.35 :rocket:

PR #45, opened 2019-06-28
fix: ProgressDot accessibility

jsomsanith:
`ProgressDot` component is not accessible
* no text alternative.
* a loader should be a live region so SR announce any change after loading.

* set an `aria-label`
* discussed with @domyen, this component is an intermediate loading state.  Set aria-live to `polite` and role to `status`.


kylesuss:
:rocket: PR was released in v0.0.34 :rocket:

PR #48, opened 2019-07-01
Memoize LinkWrapper in tooltip/ListItem for perf

kylesuss:
This has come up a few times in @jsomsanith 's PRs, but we should avoid creating these styled components in the render function as it can create a perf issue. I think this is the last one we need to update.


kylesuss:
:rocket: PR was released in v0.0.36 :rocket:

PR #50, opened 2019-07-01
fix: ListItem accessibility

jsomsanith:
`ListItem` component semantic is incorrect, using div to gather the anchors.

* use `<ul>` to wrap the list of elements
* use `<li>` to wrap each item


kylesuss:
:rocket: PR was released in v0.0.38 :rocket:

PR #52, opened 2019-07-02
Remove Tooltip onClick duplication

### LINKED ISSUE #51: Tooltip with trigger="click" behavior
**Issue**
When I click on the tooltip I can't close it unless I click outside of it.

**Expected behavior**
The tooltip should toggle between showing and hiding when I click on it.

**Gif**
Screen Recording 2019-07-02 at 12 57 PM

This is a regression because the tooltip is working as expected on Chromatic's homepage.
Screen Recording 2019-07-02 at 01 02 PM

kylesuss:
`toggleTooltipShown` was being called twice as we were manually passing an `onClick` to the tooltip trigger and the `react-popper-tooltip` library was also calling the toggle function as a result of the `onVisibilityChange`  prop.

I just remove our manual `onClick` handler and it appears to be unnecessary.

Closes #51

### DIFF
--- a/src/components/tooltip/WithTooltip.js
+++ b/src/components/tooltip/WithTooltip.js
@@ -81,7 +81,7 @@ function WithTooltip({
       )}
     >
       {({ getTriggerProps, triggerRef }) => (
-        <Container ref={triggerRef} {...getTriggerProps()} {...props} onClick={toggleTooltipShown}>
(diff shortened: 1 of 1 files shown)


kylesuss:
:rocket: PR was released in v0.0.39 :rocket:

PR #55, opened 2019-07-03
fix: Spinner accessibility

jsomsanith:
`Spinner` component is not accessible, this PR fixes the same issues as ProgressDot (#45)
* no text alternative.
* a loader should be a live region so SR announce any change after loading.

* set an `aria-label`.
* this component is used to load subparts of the page. It should not be an assertive region or it can be a hell for screen readers. Set aria-live to `polite` and role to `status`.


domyen:
:rocket: PR was released in v0.0.42 :rocket:

PR #62, opened 2019-08-07
Ignore stories in compiled output.

leerob:
### Overview

When compiling the design system for distribution via Babel, we can ignore bundling the `stories.js` files. 

I also deleted a test that didn't appear to be doing anything, but I can add that back if I'm wrong.

### Result
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   ],
   "main": "dist/index.js",
   "scripts": {
-    "build": "babel src -d dist",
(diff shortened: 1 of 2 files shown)


domyen:
Thanks so much @leerob 🙏

domyen:
Might have to roll this back. This might break the ability for consumer Storybooks to browse the design system alongside their components which is an important feature for SDS adoption (makes it easier to reuse).

leerob:
@domyen No worries - that makes sense!

PR #64, opened 2019-08-07
Integrate Rollup

kylemh:
Could somebody cut an alpha release off this so I can confirm that the changes work as expected in consumers?


.size-snapshot.json
```diff
@@ -0,0 +1,40 @@
+{
```
kylemh: Note now you'll be able to spot any massively poor changes to the library through the committed diff!

package.json
```diff
@@ -101,13 +103,24 @@
...
+    "rollup-plugin-replace": "^2.2.0",
+    "rollup-plugin-size-snapshot": "^0.10.0",
+    "rollup-plugin-sourcemaps": "^0.4.2",
+    "rollup-plugin-terser": "^5.1.1",
     "sort-package-json": "^1.21.0",
     "storybook-chromatic": "^1.3.3"
   },
   "peerDependencies": {
-    "prop-types": "^15.5.4",
-    "react": "^15.0.0 || ^16.0.0",
```
kylemh: Y'all are using `useMemo` and other more recent features, so `16.8.0` at least seems fair. Since `styled-components` doesn't eliminate itself on build-time, you're also mandating that consumers require it.

package.json
```diff
@@ -101,13 +104,25 @@
...
-    "react-dom": "^15.0.0 || ^16.0.0"
+    "polished": ">= 3.3.0",
```
kylesuss: Can you explain why `polished` needs to be a peer dep? The others seem to make sense to me, but not this one

kylesuss:
@kylemh Thanks for tackling this so quickly. Question before I review -- we are thinking of reverting this PR:

https://github.com/storybookjs/design-system/pull/62

The reason being that we like to import the stories into the storybook for the consumer application so devs know what they have access to.

With that in mind & considering your updates here, is it safe to include the stories in the build for the design system? Will they stay out of the bundles for consumer apps unless they are explicitly imported?

Here is how we typically add them to the consumer storybook fwiw:

https://github.com/storybookjs/frontpage/blob/dd764fc932f4f26e887ca17c8b23787574cecef3/.storybook/config.js#L45

domyen:
FYI Reverted #62 in #68.

package.json
kylemh: https://github.com/styled-components/polished#compatibility

In the same sense that styles have been dictated with `styled-components`, `polished` only pairs with libraries that support styling via objects. It _may_ be unnecessary (if somebody uses `styled-components` without `polished`), but typically people pair the libraries, so I figured it'd be best to place here.

kylemh:
🤔 That seems unideal to me, but you can easily do that. Why export stories? If people want access to stories from this library they should look at source (preferably via a statically hosted Storybook instance). You'd be encouraging a tight coupling of Storybook versioning next to your library's Storybook version; incompatibilities at that layer could then prevent consumption at the real consumer layer.

If a consumer has stories that consume components from this library, nothing would prevent them from defining those stories.

package.json
ghengeveld: I've personally used SC in dozens of projects, never with `polished`. So this assumption is not necessarily true. Setting it as a peer dependency practically forces people to install it, even if it's not needed. Doesn't sound right to me. Can we find a different solution?

.size-snapshot.json
ghengeveld: That's cool, but I would prefer to have a bot guard over our bundle size, rather than doing a manual review of some numbers. How is this file used?

package.json
kylemh: If you think about it though. They are installing it, the moment they install this dependency.

We can easily not add it as a peerDep and bundle it as part of this library

.size-snapshot.json
kylemh: The plugin has an option for `matchSnapshot` to enforce it in CI/CD.

https://github.com/TrySound/rollup-plugin-size-snapshot/issues/33

package.json
ghengeveld: Personally I want to get rid of reliance on such hooks in the codebase, not to be more backwards compatible but to simplify it.

kylemh: That isn't in my control, but if you convince the others' idc either way!

kylesuss:
@kylemh 

> If people want access to stories from this library they should look at source

I suppose our mentality is that there is a workflow benefit to having all the things you have access to in 1 place vs having to jump around. That and also you may be running a diff version of the design system than what is "current" in the separate deployment of the design system's Storybook.

> You'd be encouraging a tight coupling of Storybook versioning next to your library's Storybook version; 

I hear you on this. You are def right.

At this point, it sounds like a pro/con analysis of the workflow benefit to the potential for coupling issues. I do think that if we run into coupling issues in the future, it would be pretty trivial to just avoid importing the stories into Storybook so I'd rather save that concern until we have an issue!

> Don't worry about things staying out of the bundle. Everything is now tree-shakeable with this setup. You can export whatever the heck you want and it'll only add to consumer's bundle(s) if they consume it.

💥 Perfect. Love it.

kylemh:
@kylesuss sounds like the MO is export the stories!

How do you feel about me exporting them to a `/stories` folder?

If not, would you simply like them exported in the same place?

kylesuss:
@kylemh 

> How do you feel about me exporting them to a /stories folder?

Totally OK w/ this! Makes sense to separate them, I'd say.

Thanks.

kylemh:
May need some help here...

Screen Shot 2019-08-23 at 1 59 18 AM

kylemh:
Also, can I get access to Storybook NPM just to be able to cut sample releases to test the lib out with these different configs. I don't trust relative path imports across projects.

kylesuss:
@kylemh I am just getting back from some traveling & catching up -- are you able to accomplish what you need to test things with `npm link`?

https://docs.npmjs.com/cli/link

kylemh:
Definitely. Never used that before!

kylesuss:
@kylemh Nice! I've had some issues trying to get it working in the past on other projects but in theory, it is meant for this purpose. Let me know if you run into any hurdles with it and I can take a look.

As for the circular dep issue, I'm not sure I understand the problem yet. This didn't come up when we were just building w/ Babel right? What do you think is different about the new setup you are working on that is causing it?

I think it make sense to compile the stories w/ Babel if the rest of the files are doing so. Is that complicating things?

kylemh:
The circular dependency issue should be resolvable. I'm just not aware of _how_ to resolve it. I think we should reach out to Norbert or Tom on this, but the idea of transpiling the stories with babel seems like it could work.

ndelangen:
@kylemh what's the status of this, do you need help?

PR #66, opened 2019-08-08
Fix hook usage

kylemh:


src/components/tooltip/ListItem.js
```diff
@@ -164,7 +164,7 @@ export function ListItem({
   );
 
   if (LinkWrapper) {
-    const StyledLinkWrapper = useMemo(() => buildStyledLinkWrapper(LinkWrapper), [LinkWrapper]);
+    const StyledLinkWrapper = buildStyledLinkWrapper(LinkWrapper);
```
kylesuss: There is another usage similar to this:

https://github.com/storybookjs/design-system/blob/a46acaa615164ed1e7ad650ba92a9f6d717ec248/src/components/Button.js#L335

Do you think it needs updating as well?

kylemh: I asked on dischord if any consumers are using SSR. It seems like no usage of `useMemo` is working in a Next.js application. I'm gonna see how it all goes in a CRA environment today.

kylesuss:

PR #68, opened 2019-08-08
Revert "Ignore stories in compiled output."

domyen:
Breaks the ability for consumer Storybooks to browse the design system alongside their components which is an important feature for SDS adoption.


leerob:
Might want to gut out that unused test later, but that can be a separate PR if needed.

Issue #54, opened 2019-07-02
Remove font loading in `GlobalStyle`


kylesuss:
Remove the font loading of Nunito Sans [here](https://github.com/storybookjs/design-system/blob/9bcfa744617c4fe59d001db956cbc3139ec88aca/src/components/shared/global.js#L115) or potentially configure the `GlobalStyle` to conditionally load it.

Currently, design system consumer app frameworks could re-render layout components (Gatsby is an example of a framework that does this) on each page change. This causes the font to be requested on every page change and also a flickering of fonts when that happens. 

It might be easiest to just pass a prop to `GlobalStyle` like this:

```javascript
export const GlobalStyle = createGlobalStyle`
  ${props => !props.withoutFonts && `
    @import url('https://fonts.googleapis.com/css?family=Nunito+Sans:400,700,800,900');
  `

  body {
    ${bodyStyles}
    margin: 0;
    overflow-y: auto;
    overflow-x: hidden;
  }
`;

// Should we assume that font loading is the default?
<GlobalStyle />
<GlobalStyle withoutFonts />
```

...but other ideas are welcomed as well.

qoalu:
This wouldn’t fix the problem of rerendering (reloading and flicker), right?

kylesuss:
@qoalu no def not. My thought process was that the app consuming the design system would have to then setup their font loading manually in a way that made sense for their dev environment/framework/library. This was just a way to give them the opportunity to do that.

qoalu:
I was contemplating a similar approach in our design system. Would be interested to see what other options people come up with.

Another idea is to make a side-effect, somehow outside the react lifecycle

kylesuss:
Nice. What do you mean by side-effect?

qoalu:
Something like useEffect and creating a node with a certain id in the dom.
Keeping track if it not to post the same code twice.
Emotion 9 has an injectGlobalStyle that just hard pasted code into html, actually

kristof0425:
I did some research 🧐on your problem and I've found the following:

There was a long discussion about the same problem you experience at [styled-components](https://github.com/styled-components/styled-components/issues/1593).

The gist of it is the following:
- `createGlobalStyle` won't have any future release, which is going to solve this problem, as they've found the solution is out of `styled-components` context, the maintainers locked this issue

Possible solutions were suggested:
- Tweak the font files caching headers, so the browser won't re-request them. (I suppose this is not an option for you, as you have no control over the caching headers in this case.)
[...]

PR #76, opened 2019-09-11
Move remaining body styles to `bodyStyles`

tmeasday:
This isn't a complete fix for #54, but it is a step forward, as it allows users to safely do:

```js
import { createGlobalStyle } from 'styled-components';
import { global } from '@storybook/design-system';

const CustomGlobalStyle = createGlobalStyle`
   body {
     ${global.bodyStyles}
   }
}`
```

In their own code to avoid the `@import`.


kylesuss:
:rocket: PR was released in v1.0.0 :rocket:

PR #77, opened 2019-09-11
Remove font import from GlobalStyle, export fontHref

### LINKED ISSUE #54: Remove font loading in `GlobalStyle`

kylesuss:
Let's just leave the font import out of the GlobalStyle and export the fontHref so that the design system consumers can share font settings, but load them in the most appropriate way for their environment.

Closes #54

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,6 +1,7 @@
 import React from 'react';
 import { configure, addDecorator } from '@storybook/react';
 import { withA11y } from '@storybook/addon-a11y';
+import { fontHref } from '../src/components/shared/global';
 import 'storybook-chromatic';
 
 import { GlobalStyle } from '../src/components/shared/global';
@@ -21,3 +22,16 @@ configure(
   ],
   module
 );
+
+// Load the font and avoid re-loading it when components change
+const fontLinkId = 'font-link';
+
+if (!document.getElementById(fontLinkId)) {
+  const fontLink = document.createElement('link');
+
+  fontLink.id = fontLinkId;
+  fontLink.href = fontHref;
+  fontLink.rel = 'stylesheet';
+
+  document.head.appendChild(fontLink);
+}
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import { configure, addDecorator } from '@storybook/react';
 import { withA11y } from '@storybook/addon-a11y';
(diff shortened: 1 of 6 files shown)


kylesuss:
@domyen yep. Need a bit as well. I pulled the trigger too early. I have to load the font in Storybook here, which should help w/ showcasing how to use it. Then I can address those points!

.storybook/config.js
```diff
@@ -21,3 +22,16 @@ configure(
   ],
   module
 );
+
+// Load the font and avoid re-loading it when components change
```
domyen: If we're going to load this in every Storybook config I wonder if we should make this a function that is exported with the design system?

`withStorybookDSFont` or something.

kylesuss: Yeah I thought about that but it felt like a slippery slope -- aka are we going to do that for every framework that we use? Should the design system make assumptions about that kinda stuff? I suppose with Storybook that is safe. Just thinking out loud.

kylesuss: https://github.com/storybookjs/design-system/pull/77/commits/ed8aafa3dbed5c1b771af067b5119fea32615fd3

domyen:
lgtm, might want to check on the build failure though

kylesuss:
Build failure seemed transient & unrelated to the changes in this PR.

kylesuss:
:rocket: PR was released in v1.0.0 :rocket:

PR #79, opened 2019-09-25
Add global styling setup to README

kylesuss:
All of the client applications that use this design system depend on the global styling that we export, thought it has remained undocumented.  This PR adds the instructions on how to use the global styling to the `README`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ Components within the design system assume that a set of global styles have been
 
 #### Option 1: Render the `GlobalStyle` component
 
-Useful when you don't need any custom `body` styling in the application.
+Useful when you don't need any custom `body` styling in the application, typically this would be placed in a layout component that wraps all pages, or a top-level `App` component.
 
 ```javascript
 import { global } from '@storybook/design-system';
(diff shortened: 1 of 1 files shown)


kylesuss:
@tmeasday you have touched this recently in the consumer apps so are probably the most qualified to review, if possible? 

https://github.com/storybookjs/design-system/blob/947b4322c8ac672116a27e651398619dec2c146f/README.md#global-styles

README.md
```diff
@@ -44,18 +44,44 @@ The Storybook design system codifies existing UI components into a central, well
...
-## Usage
+## Global Styles
 
-```jsx
-import React, { Component } from 'react';
+Components within the design system assume that a set of global styles have been configured. Depending upon the needs of the application, this can be done several ways:
+
+#### Option 1: Render the `GlobalStyle` component
+
+Useful when you don't need any custom `body` styling in the application.
```
tmeasday:
```suggestion
Useful when you don't need any custom `body` styling in the application, typically this would be placed in a layout component that wraps all pages, or a top-level `App` component.
```

Maybe a small hint about where to put it might help.

kylesuss:
@tmeasday sounds good. Merging this for now & filed a ticket for the enhancement:

https://github.com/storybookjs/design-system/issues/83

kylesuss:

PR #85, opened 2019-10-07
Add delayHide prop to tooltip for better hover behavior

### LINKED ISSUE #82: Fix tooltip hover issue
Sometimes the tooltip can have issues where when you try to hover over the dropdown, it disappears instantly:

tooltip

Maybe the hover target is too small?

kylesuss:
Thanks for the idea @domyen . Def seemed to do the trick.

Before:
tooltip-before

After:
tooltip-after

Closes #82


kylesuss:

PR #86, opened 2019-10-07
Fix Highlight PropType error for CSS langauge

kylesuss:
Was getting a PropType error for this. Missed one of the languages in the Highlight component refactor.


kylesuss:

PR #87, opened 2019-10-08
Add back global styles for code, pre elements

kylesuss:
I moved some global styles around in #80 that had a negative effect when importing the global styles in the consumer apps. Adding them back here.


kylesuss:
:rocket: PR was released in v1.1.3 :rocket:

PR #88, opened 2019-10-08
Remove default styling for Highlight code, pre selectors, conflict with globals

kylesuss:
After a bit of debugging, I realize that the "theme" I brought over for prism js was overriding a bunch of our global styles for code & pre elements in `Highlight` blocks. I think it is best to just remove those styles and instead default to our own global styling. I have left the theme styles for prism that target highlighted elements of the code.

### DIFF
--- a/src/components/Highlight.js
+++ b/src/components/Highlight.js
@@ -5,7 +5,7 @@ import styled from 'styled-components';
 import Prism from 'prismjs';
 import loadLanguages from 'prismjs/components/index';
 import PropTypes from 'prop-types';
-import { color, typography } from './shared/styles';
+import { color } from './shared/styles';
 
 const languages = ['bash', 'javascript', 'typescript', 'json', 'css'];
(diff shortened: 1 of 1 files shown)


kylesuss:
@domyen there were some changes that I approved on Chromatic. They were caused by:

1. Removing the line-height inconsistency in the Prism JS theme. For example:

https://www.chromaticqa.com/snapshot?appId=5ccbc373887ca40020446347&id=5d9cc25d9bdaca0020abeff1

2. Removing a style from the Prism theme that affected how we style certain things in Bash. For example:

_Note: this one also has the line height diff_

https://www.chromaticqa.com/snapshot?appId=5ccbc373887ca40020446347&id=5d9cc25d9bdaca0020abeff3

domyen:
This looks good to me. Just inspected in Chromatic live view.

kylesuss:
:rocket: PR was released in v1.1.4 :rocket:

PR #89, opened 2019-10-08
Add left, right padding back to Highlight component

kylesuss:
Looks like the consumers are actually utilizing the global padding on the `code` blocks.


domyen:
Nice!

kylesuss:
:rocket: PR was released in v1.1.5 :rocket:

PR #90, opened 2019-10-08
Fix wrapping, overflow on Highlight component

kylesuss:
...to avoid having code break to the next line when the line is long.


kylesuss:
:rocket: PR was released in v1.1.6 :rocket:

PR #95, opened 2019-10-17
Replace custom colors/typography stories with DocBlocks.

leerob:
Partially addresses https://github.com/storybookjs/design-system/issues/84.

## Overview

With official MDX support coming soon, we want to update the design system to reflect the new hotness. This PR updates the custom colors/typography stories to use MDX along with the corresponding DocBlocks.

## Examples

Screen Shot 2019-10-16 at 8 52 33 PM
Screen Shot 2019-10-16 at 8 52 43 PM

PR #96, opened 2019-10-17
Add story names to preserve component history

kylesuss:
I noticed in #95 that a large number of the stories had different names -- they were capitalized and spaces were added. As it turns out, Storybook uses `startCase` when you use CSF. We missed out on that originally as we were on a prerelease. I think I'd like to preserve the component history on Chromatic & therefore would opt to define the story names manually for now.

@leerob I am hoping to merge this into your `typography-and-colors` branch so that we can have Chromatic run again and only isolate the changes from your effort in there.

### DIFF
--- a/src/components/Avatar.stories.js
+++ b/src/components/Avatar.stories.js
@@ -22,6 +22,10 @@ export const large = () => (
   </div>
 );
 
+large.story = {
+  name: 'large',
+};
+
 export const medium = () => (
   <div>
     <Avatar loading />
@@ -29,6 +33,10 @@ export const medium = () => (
     <Avatar username="Tom Coleman" src="https://avatars2.githubusercontent.com/u/132554" />
   </div>
 );
+
+medium.story = {
+  name: 'medium',
+};
 
 export const small = () => (
   <div>
@@ -42,6 +50,10 @@ export const small = () => (
   </div>
 );
 
+small.story = {
+  name: 'small',
+};
+
 export const tiny = () => (
   <div>
     <Avatar loading size="tiny" />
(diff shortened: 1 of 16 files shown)


kylesuss:
@shilman does this make sense to you? I made this PR to try and control the names of the stories & Chromatic is still showing the capitalized names 🤔

tmeasday:
Ok, took a look at this @kylesuss. A couple of things.

1. The reason for original problem was due to the design system being on a beta of 5.2, and the feature to capitalize story names from exports hadn't yet come in.

2. We should probably just lean into that now and wear the chromatic "new stories" (chromatic can't track stories across renames) as part of the upgrade to 5.2 proper. So *not* do this PR.

3. The feature that you used to rename the stories isn't currently supported by chromatic (https://github.com/chromaui/chromatic/issues/2549) so that's why they still appear capitalized in Chromatic (its show the `story.name` rather than the `story.displayName`).

4. Chromatic actually would still treat these stories as new, even with the new `displayName` as the underlying `name` is what we use to track stories. This seems like the right call and lets you rename stories (as long as you don't change the export name) without losing track of them in Chromatic.

kylesuss:
Thanks @tmeasday. Slightly confusing thing about what you are saying:

> [Chromatic] shows the story.name rather than the story.displayName

Bit of a tricky one to wrap my head around since I am doing this:

```javascript
derp.story = {
  name: 'derp',
}
```

vs 

```javascript
derp.story = {
  displayName: 'derp',
}
```

Maybe under the hood the `name` that I am passing is getting remapped to `displayName`? That is how I read your comment.

Ultimately, the takeaway here sounds like we should just roll with this and let the display names be generated automatically.

tmeasday:
I agree that’s confusing but you have it right. @shilman do you remember if we thought about the fact that that name is super confusing? I propose we change it to `displayName` and deprecate the `name` option.

PR #122, opened 2019-11-28
Rename 'loading' prop to 'isLoading' to avoid unknown attribute warnings

ghengeveld:
Also fixes a mistake in the intro page.


ndelangen:
I think releasing this requires a major version bump.

PR #155, opened 2020-02-27
Render tooltip portal in #root if it exists

ghengeveld:


ghengeveld:
:rocket: PR was released in v4.0.0 :rocket:

PR #156, opened 2020-03-01
Downgrade Styled Components back to v4

ghengeveld:
Styled Components v5 has a [memory leak](https://github.com/styled-components/styled-components/issues/3022) and [occasionally fails](https://github.com/styled-components/styled-components/issues/3030) in `createGlobalStyle`.


ghengeveld:
:rocket: PR was released in v5.0.0 :rocket:

wwwebman:
This makes me said, we have to do the same in our code - downgrade, meh :(

PR #160, opened 2020-03-21
Refine outline and inverse button styling for :hover and :focus states

domyen:
Made subtle changes to the Button's outline and inverse states for :focus and :hover. The outline state now works better on colored backgrounds. The inverse states now get a focus state that corresponds to their colors (pink and blue).

This video shows the new states.
https://www.loom.com/share/d87ea2545696464389d38cf66277eee1

@kylesuss Can you release and update when you get a chance? There are no changes in Chromatic because the rest states should be identical.


kylesuss:
:rocket: PR was released in v5.0.4 :rocket:

PR #161, opened 2020-03-27
Make border-radius more consistent

domyen:
Use border-radius variable from shared styles instead of an absolute value.

In practice, this changes the border radius of some elements from 4px to 5px.


ndelangen:

PR #162, opened 2020-04-05
Remove keycode, reduce bundle size

oliviertassinari:
A simple win I noticed looking at the source of the design system. You can learn more about how it can work in https://github.com/mui-org/material-ui/pull/14248 (it only works with synthetic events).

### DIFF
--- a/package.json
+++ b/package.json
@@ -45,7 +45,6 @@
     ]
   },
   "dependencies": {
-    "keycode": "^2.2.0",
     "polished": "^3.4.1",
     "prismjs": "1.17.1",
     "react-modal": "^3.10.1",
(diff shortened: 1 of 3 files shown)


oliviertassinari:
I'm looking into how we can increase the usage of the Material-UI components in the OSS community, while the unstyled and style engine interop [stories](https://github.com/mui-org/material-ui/issues/20012) we have for v5 is meant to greatly help, we might not get it stable before early 2021. For now, I'm looking at existing opportunities that can benefit everybody as well as learning more about the use cases.

If you are interested, I have noticed that we can win 1.6 kb with a change of the Modal: [6.5 kB](https://bundlephobia.com/result?p=react-modal) -> [4.9 kB](https://bundlephobia.com/result?p=@material-ui/core).

Capture d’écran 2020-04-05 à 16 48 56

domyen:
Thanks @oliviertassinari will take a deeper look in the next couple of days. Appreciate the link to the MUI issue 🙏

ndelangen:

PR #171, opened 2020-06-23
Use github secrets for new project token

kylesuss:
The project token has been updated already & set as a secret in this repo. Access it at Chromatic if needed.


kylesuss:
:rocket: PR was released in v5.1.1 :rocket:

PR #172, opened 2020-07-07
Remove recompose

kylesuss:
An unnecessary dependency at this point.


kylesuss:
:rocket: PR was released in v5.1.1 :rocket:

PR #173, opened 2020-07-14
Update styled-components dependency

domyen:
I tried to reproduce a styled-components issue by updating the dependency.  https://github.com/storybookjs/storybook/issues/9523#issuecomment-631337063

It doesn't seem like I can reproduce the issue, but it's useful to update the dep anyways.

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "prop-types": "^15.5.4",
     "react-modal": "^3.11.2",
     "react-popper-tooltip": "^2.11.1",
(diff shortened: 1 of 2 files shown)


kylesuss:
:rocket: PR was released in v5.1.2 :rocket:

kylesuss:
@domyen I forgot about this, but we can't update this yet as a result of:
https://github.com/styled-components/styled-components/issues/3022

I'll migrate back for now.

PR #176, opened 2020-07-21
Do not dynamically generate styled link component in render

### LINKED ISSUE #174: Incorrect use of `styled`
This commit: https://github.com/storybookjs/design-system/commit/6bed38726e796ad15ef47805b56b342fc21a8f2e

Introduces a memory leak as you are definitely not supposed to use `styled` inline inside a component. What that does is create a leak where a new style definition is created every time you render the component.

kylesuss:
The solution to this needs to:

1. Allow a link to have a `LinkWrapper`
2. Allow a link to be a button (that looks like a link)
3. Not pass down arbitrary props to an `a` element (like `nochrome` or `inverse`) as that throws console warnings
4. Not dynamically generate styled components during render

Think I have all 4 covered!

Closes #174

### DIFF
--- a/src/components/Link.js
+++ b/src/components/Link.js
@@ -10,7 +10,6 @@ const linkStyles = css`
   display: inline-block;
   transition: transform 150ms ease-out, color 150ms ease-out;
   text-decoration: none;
-
   color: ${color.secondary};
 
   &:hover,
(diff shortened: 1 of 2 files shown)


kylesuss:
@tmeasday 

> LGTM -- one thing though: this assumes a LinkWrapper generates an <a> right? Is that something we've always guaranteed?

No I don't think we have guaranteed that, but I am struggling to think of a time where that wouldn't be true. What use case are you thinking of?

tmeasday:
> No I don't think we have guaranteed that, but I am struggling to think of a time where that wouldn't be true. What use case are you thinking of?

No idea, I was just checking! If you think it is good, let's release this!

kylesuss:
:rocket: PR was released in v5.1.2 :rocket:

PR #177, opened 2020-07-23
Do not dynamically generate styled button component in render

kylesuss:
Basically a dup of #176 but for `Button`s this time (I forgot we did the same thing here).


kylesuss:
:rocket: PR was released in v5.1.3 :rocket:

PR #178, opened 2020-07-24
Use 'as' prop to control button, link styles and avoid transient props

kylesuss:


kylesuss:
:rocket: PR was released in v5.1.4 :rocket:

PR #180, opened 2020-07-30
Table of Contents component

kylesuss:
A highly dynamic table of contents component to be used in the consumers. The challenges w/ this component were mostly related to state management given the following constraints:

* The TOC can have a random combination of different types of children (I called them "menu", "link", and "bullet link")
* The TOC should be able to have the state of all of its children controlled at once externally (a button to expand/collapse all)
* Changing pate path activates menu items that were not previously open
* Keyboard navigation opens menu items that were not previously open so you can tab through the entire menu


kylesuss:
@domyen sure thing. I think it was one of those things I thought belonged together at the beginning because you'll never use them individually, but agreed that it got too big. I ended up separating more in this way:

1. State management
2. List
3. List items

There is 1 caveat to that which is 1 of the list items is a recursive component render, so it is not broken out further (it lives in `TableOfContentsItems`).

src/components/TableOfContents.js
```diff
@@ -112,20 +116,18 @@ BulletLink.propTypes = {
...
-  margin-bottom: 12px;
 `;
 
 const MenuLink = styled(({ isActive, ...rest }) => <Link {...rest} />)`
   color: ${(props) => (props.isActive ? color.secondary : color.darkest)};
   font-weight: ${(props) => (props.isActive ? typography.weight.bold : typography.weight.regular)};
-  margin-bottom: 12px;
 `;
 
 const MenuChild = styled.div`
```
domyen: This drops an extra `div` into the markup. Can we consolidate the styles into the `ul`? 
image

kylesuss: Yeah sure. I think the need for this is gone after the refactor to use ul & ol. Update here:

https://github.com/storybookjs/design-system/pull/180/commits/401233c1e1c4bd6209eecdae3b0e543a94587fa4

domyen:
lgtm, thanks @kylesuss

kylesuss:
:rocket: PR was released in v5.1.6 :rocket:

PR #183, opened 2020-08-03
Pass top level menu state to consumers of the TOC component

kylesuss:


kylesuss:
:rocket: PR was released in v5.1.8 :rocket:

PR #184, opened 2020-08-03
Export styled button from ButtonComponent

kylesuss:
...so that it can be used with the `as` prop for styled components.


kylesuss:
:rocket: PR was released in v5.1.9 :rocket:

PR #185, opened 2020-08-03
Port over ShadowBoxCTA from learnstorybook.com

kylesuss:
A direct port of the component & stories from LSB.


kylesuss:
:rocket: PR was released in v5.1.10 :rocket:

PR #191, opened 2020-08-07
Fix shadow box CTA required props

kylesuss:


kylesuss:
:rocket: PR was released in v5.1.16 :rocket:

PR #195, opened 2020-08-10
Minor refinements to codesnippet

domyen:


src/components/CodeSnippets.js
```diff
@@ -28,8 +28,8 @@ const StyledHighlight = styled(Highlight)`
 
   && pre[class*='language-'] {
     background-color: transparent;
-    padding: 15px 1rem;
-    padding-right: 5rem;
+    padding: 1em 1em 1.5em;
+    padding-right: 2em;
```
kylesuss, line 32: The point of this was to make it so that you could scroll all the way right & not have the "Copy" button overlaying any of the code. So long as you are aware that removing this will break that, I am ok w/ it. Maybe the argument is that you can see it when you are scrolled left?

domyen: The internal padding shouldn't trigger the scrollbar unnecessarily.

afaict the scrollbar appears below the content so it doesn't affect the content or the copy functionality.

image

kylesuss: > The internal padding shouldn't trigger the scrollbar unnecessarily.

That's fair, though it has the potential to overlap slightly: 
Screen Shot 2020-08-10 at 11 24 20 AM

src/components/CodeSnippets.js
```diff
@@ -137,7 +138,7 @@ export function CodeSnippets({ snippets, ...rest }) {
   const [activeSnippetIndex] = useState(0);
 
   return (
-    <Wrapper {...rest}>
+    <Wrapper {...rest} className="codesnippet">
```
kylesuss, line 141: You can target this w/ `styled(CodeSnippets)`

domyen: For docs formatting we use a styled `css` helper which doesn't support `styled()`.

image

kylesuss: The code snippets work a bit differently. We already have a styled code snippet, but its kinda hard to track down. Try editing this:

https://github.com/storybookjs/frontpage/blob/next/src/components/screens/DocsScreen/CodeSnippets.tsx#L31

kylesuss: If instead you'd like to keep the styles together, that's fine. Just trying to shed some light on how the code snippets work.

kylesuss: Ultimately, this new MDX method we have means we can avoid this kinda of css based styling targeting selectors in later projects. We can supply components for each of the tags, but that is probably something to work through another day.

src/components/CodeSnippets.js
domyen: I can live with this 👍 
image

kylesuss:
:rocket: PR was released in v5.1.19 :rocket:

PR #199, opened 2020-08-10
Setup slot in the CodeSnippets for non-code things (like messages)

kylesuss:


kylesuss:
:rocket: PR was released in v5.1.22 :rocket:

PR #201, opened 2020-08-11
Simplify release process

kylesuss:


scripts/update-chromatic-url.ts
```diff
@@ -1,26 +0,0 @@
...
-    },
-  };
-
-  await writeJson(packageLocation, contents, { spaces: 2 });
-};
-
-run().catch(e => {
-  console.error(e);
-  process.exit(1);
-});
```
kylesuss, line 26: This seems like overkill given wen can just use this: https://master--5ccbc373887ca40020446347.chromatic.com

kylesuss:
:rocket: PR was released in [`v5.1.24`](https://github.com/storybookjs/design-system/releases/tag/v5.1.24) :rocket:

scripts/update-chromatic-url.ts
kylesuss: Also this script was making `auto` stop working because it created an unclean working directory (did not commit the changes to package.json). Seems to just complicate things at this point, hence the removal.

Issue #159, opened 2020-03-18
Pass Forwarding Refs to the component


leonardoelias:
Today the components are like this:

```
function Button({
  isDisabled,
  isLoading,
  loadingText,
  isLink,
  children,
  ButtonWrapper,
  ...props
}) {
  // ... rest components

  return (
    <SelectedButton isLoading={isLoading} disabled={isDisabled} {...props}>
      {buttonInner}
    </SelectedButton>
  );
}
```

to 

```
const Button = React.forwardRef(
  (
    {
      isDisabled,
      isLoading,
      loadingText,
      isLink,
      children,
      ButtonWrapper,
      ...props
    },
    ref
  ) => {
    // ... components

    return (
      <SelectedButton
        ref={ref}
        isLoading={isLoading}
        disabled={isDisabled}
        {...props}
      >
        {buttonInner}
      </SelectedButton>
    );
  }
);
```

So we can  we want to forward the ref from the outside of.
https://reactjs.org/docs/forwarding-refs.html

I can work on that refctory and open a PR for better evaluation.

However, I would like to know how to trade.

kylesuss:
Hi 👋 . Thanks for the suggestion. Can you describe the use case that you think the ref needs to be forwarded? So far we haven't needed that, but maybe I am missing something useful.

PR #204, opened 2020-08-18
Forward ref for button, link components

kylesuss:
Partial implementation of #159 

I am basically adding this because of:

https://github.com/vercel/next.js/issues/7915

Now all function children of next's `Link` component need to forwardRefs. I am keeping it to just the `Button` and `Link` for now as that is likely all that is a child of a link. Happy to upgrade other components if necessary, but don't want to over-optimize.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { forwardRef } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


kylesuss:
:rocket: PR was released in [`v5.1.26`](https://github.com/storybookjs/design-system/releases/tag/v5.1.26) :rocket:

PR #211, opened 2020-10-06
includes className prop on tooltip/TooltipMessage component

D3Portillo:
This commit also includes an example of how to mod component using styled components.

### DIFF
--- a/src/components/tooltip/TooltipMessage.js
+++ b/src/components/tooltip/TooltipMessage.js
@@ -32,9 +32,9 @@ const MessageWrapper = styled.div`
   box-sizing: border-box;
 `;
 
-export function TooltipMessage({ title, desc, links, ...rest }) {
+export function TooltipMessage({ title, desc, links, className, ...rest }) {
   return (
-    <MessageWrapper {...rest}>
+    <MessageWrapper className={className} {...rest}>
       <Message>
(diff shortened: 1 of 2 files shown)


kylesuss (changes requested):
Hey @D3Portillo , I may have misspoken slightly. I just noticed this commit:
https://github.com/storybookjs/design-system/commit/99442ccb8cd5974adb0fa5bfb960d2644e55571c

I think perhaps this may have been added and the original issue wasn't closed. That said, I think the story that you added here is valuable. Can you do me a favor? Try removing your changes in `TooltipMessage.js` and see if your story still renders the same. I suspect it will because we are passing down `...rest` to the `MessageWrapper`. That's typically the pattern we follow.

Would love to merge your story in once you confirm!

🙇

D3Portillo:
Will do, no problem : )

D3Portillo:
Yes it renders as expected. I tough you needed a prop validation or the `className` prop to be explicit in there : ) .  Deleting changes...

kylesuss (approved):
Thanks for this. Sorry I lead you a bit astray. This story is still quite useful!

kylesuss:
:rocket: PR was released in [`v5.1.27`](https://github.com/storybookjs/design-system/releases/tag/v5.1.27) :rocket:

PR #212, opened 2020-10-15
TagItem and TagList components

winkerVSbecks:
Not sure if this is the best API for the `TagList`

### DIFF
--- a/src/components/tag/TagItem.stories.js
+++ b/src/components/tag/TagItem.stories.js
@@ -1,0 +1,12 @@
+import React from 'react';
+import { TagItem } from './TagItem';
+import { StoryLinkWrapper } from '../StoryLinkWrapper';
+
+export default {
+  component: TagItem,
+  title: 'Design System/Tag/TagItem',
+};
+
+export const Default = () => <TagItem>⚛️ React</TagItem>;
+
+export const WithLinkWrapper = () => <TagItem LinkWrapper={StoryLinkWrapper}>⚛️ React</TagItem>;
--- a/src/components/tag/TagItem.stories.tsx
+++ b/src/components/tag/TagItem.stories.tsx
@@ -1,5 +1,9 @@
 import React from 'react';
+// @ts-ignore
+import seedrandom from 'seedrandom';
 import { TagItem } from './TagItem';
+
+seedrandom('chromatic testing', { global: true });
 
 export default {
   component: TagItem,
@@ -7,7 +11,5 @@ export default {
 };
 
 export const Default = () => <TagItem>⚛️ React</TagItem>;
-
-export const WithLinkWrapper = () => <TagItem>⚛️ React</TagItem>;
 
 export const Loading = () => <TagItem isLoading />;
--- a/src/components/tag/TagItem.stories.tsx
+++ b/src/components/tag/TagItem.stories.tsx
@@ -1,6 +1,5 @@
 import React from 'react';
 import { TagItem } from './TagItem';
-import { StoryLinkWrapper } from '../StoryLinkWrapper';
 
 export default {
   component: TagItem,
@@ -9,4 +8,6 @@ export default {
 
 export const Default = () => <TagItem>⚛️ React</TagItem>;
 
-export const WithLinkWrapper = () => <TagItem LinkWrapper={StoryLinkWrapper}>⚛️ React</TagItem>;
+export const WithLinkWrapper = () => <TagItem>⚛️ React</TagItem>;
+
+export const Loading = () => <TagItem isLoading />;
--- a/src/components/tag/TagItem.tsx
+++ b/src/components/tag/TagItem.tsx
@@ -1,0 +1,52 @@
+import React from 'react';
+import styled from 'styled-components';
+import { rgba } from 'polished';
+import { color, typography, background, spacing } from '../shared/styles';
+// @ts-ignore
+import { Link } from '../Link';
+
+type TagItemProps = {
+  containsIcon?: boolean;
+  inverse?: boolean;
+  isButton?: boolean;
+  LinkWrapper?: React.ComponentType | (() => React.ReactNode);
+  nochrome?: boolean;
+  secondary?: boolean;
+  tertiary?: boolean;
+  withArrow?: boolean;
+};
+
+export const TagItem = styled(Link)<TagItemProps>`
+  background: ${background.app};
+  border-radius: ${spacing.borderRadius.small}px;
+  padding: 5px 10px;
+  font-size: ${typography.size.s2}px;
+  line-height: ${typography.size.m2}px;
+  position: relative;
+  color: ${color.darkest};
+  border-width: 1px;
+  border-style: solid;
+  border-color: transparent;
+  white-space: nowrap;
+
+  &:after {
+    content: '';
+    position: absolute;
+    z-index: -1;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+    box-shadow: rgba(0, 0, 0, 0.08) 0 3px 10px 0;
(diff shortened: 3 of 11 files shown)


src/components/tag/TagItem.tsx
```diff
@@ -0,0 +1,52 @@
...
+type TagItemProps = {
+  containsIcon?: boolean;
+  inverse?: boolean;
+  isButton?: boolean;
+  LinkWrapper?: React.ComponentType | (() => React.ReactNode);
+  nochrome?: boolean;
+  secondary?: boolean;
+  tertiary?: boolean;
+  withArrow?: boolean;
+};
```
kylesuss, line 17: These are the link props right? Is it worth defining them over there & then importing them in here so we don't duplicate them eventually?

winkerVSbecks: Was thinking that. The Link is still JS and not TS. Maybe I should switch that over first and then continue with this. Would you consider that a breaking change?

src/components/tag/TagList.stories.js
```diff
@@ -0,0 +1,127 @@
...
+  excludeStories: /.*Data$/,
+  title: 'Design System/Tag/TagList',
+};
+
+export const Default = () => (
+  <TagList
+    tags={mockTagsData.slice(0, 4).map((tag) => (
+      <TagItem key={tag.link} href={tag.link}>
+        {tag.name}
+      </TagItem>
```
kylesuss, line 16: I'm wondering if the `TagList` should automatically render the `TagItem` 🤔 

Mostly saying that as it mirrors the patterns we have so far, like:

https://github.com/storybookjs/design-system/blob/5bbe65d0bcb0b1526f39bc180c40d64f53d7be67/src/components/LinkTabs.js#L56-L60

That said, I'm open to discussion on it for sure. What was your thought process with this API?

winkerVSbecks: My intention was to avoid prop drilling. Since, the `TagItem` is a link it can be a regular link or Gatsby Link or Next Link. Which means we would have to pass down a linkWrapper and maybe even have logic around `to` vs `href`. The Tag itself could have an icon, etc. I find this wrapper approach avoids a lot of those headaches and offers a more flexible API.

src/components/tag/TagItem.tsx
kylesuss: That's right. Makes sense then. I'd say leave it if refactoring the Link component is a lot of work at this point. Refactoring it won't break the implementation by any consumers though, right? If that's the case, I wouldn't consider it breaking.

src/components/tag/TagList.stories.js
kylesuss: Yeah that's true. Makes sense. I have run into headaches w/ one of our components that renders links & has the old style prop API (not what you did here) for similar reasons. If for some reason you need to target the underlying link (or tag in this case), it becomes pretty complicated to do that. I'm 👍  w/ this approach -- maybe @domyen has some thoughts as I think he had a hand in some of those original APIs.

kylesuss (approved):
I'm good on this, pending any API comments from others 👍

src/components/tag/TagList.stories.js
domyen: I'll defer to y'all to make the technical choices. The APIs I designed were intended for a very narrow use case (the Chromatic app) – I wouldn't hold them up as best practice for supporting several apps. 🙇‍♂️

src/components/tag/TagItem.stories.js
```diff
@@ -0,0 +1,12 @@
...
+import { StoryLinkWrapper } from '../StoryLinkWrapper';
+
+export default {
+  component: TagItem,
+  title: 'Design System/Tag/TagItem',
+};
+
+export const Default = () => <TagItem>⚛️ React</TagItem>;
+
+export const WithLinkWrapper = () => <TagItem LinkWrapper={StoryLinkWrapper}>⚛️ React</TagItem>;
```
domyen, line 12: - Add a TagItem that isn't a link (we have them on LSB). It looks the same but has no hover behavior.

- Include a dummy url in TagItems that are links so that the behavior is evident. e.g., for regular links include `href="https://chromatic.com"` and for LinkWrapper include `to="https://chromatic.com"`

domyen (changes requested):

src/components/tag/TagList.stories.js
```diff
@@ -0,0 +1,127 @@
...
+    ))}
+  />
+);
+
+export const WithMoreTags = () => (
```
domyen, line 21: Now would be a good time to use Args

src/components/tag/TagList.stories.js
```diff
@@ -0,0 +1,127 @@
...
+  <TagList
+    tags={mockTagsData.map((tag) => (
+      <TagItem key={tag.link} href={tag.link}>
+        {tag.name}
+      </TagItem>
+    ))}
+    limit={6}
+  />
+);
+
```
domyen, line 41: Should this also include an example with LinkWrapper? We document it in TagItem, but I'd overlook that functionality if it's not called out in a story. 

e.g., I wouldn't know I could use `to="url"` in tandem with `LinkWrapper={StoryLinkWrapper}`

src/components/tag/TagItem.tsx
winkerVSbecks: I opened another PR converting links to TS. Will update once that is merged.

winkerVSbecks:
@domyen @kylesuss updated with requested changes. I ended up creating a TagLink to handle links. TagItem is the non-interactive version.

kylesuss:
@winkerVSbecks nice, sounds and looks good to me.

Out of curiosity, Varun, do you think all components in this design system should export themselves as a `forwardRef`? Is that a good goal to work toward? TBH, it feels weird to me to do that but also annoying when you finally end up needing that ref and have to go back and add it.

winkerVSbecks:
Yea, that seems to be the pattern most DS adopt. Makes sense since they are mostly low level components. That said, typing those components is a pain. 

I don't plan on rewriting everything. We can do that for new components and update old components as and when necessary.

jonniebigodes:
@winkerVSbecks just checked the latest commit you've made and it looks really great. I'm good with it as well 👍

src/components/tag/TagItem.stories.tsx
```diff
@@ -0,0 +1,13 @@
...
+import { TagItem } from './TagItem';
+
+export default {
+  component: TagItem,
+  title: 'Design System/Tag/TagItem',
+};
+
+export const Default = () => <TagItem>⚛️ React</TagItem>;
+
+export const WithLinkWrapper = () => <TagItem>⚛️ React</TagItem>;
```
domyen, line 11: Is there supposed to be a LinkWrapper around this?

src/components/tag/TagLink.tsx
```diff
@@ -0,0 +1,75 @@
...
+  nochrome?: boolean;
+  secondary?: boolean;
+  tertiary?: boolean;
+  withArrow?: boolean;
+};
+
+export const TagLink = styled(Link).attrs<TagLinkProps>(({ isLoading, children }) => ({
+  children: isLoading ? 'Loading tag' : children,
+}))<TagLinkProps>`
+  display: inline-block;
```
domyen, line 24: Can we extend TagItem styles instead of repeating them?

domyen (changes requested):

domyen:
Heads up, I'm leaving specific UI comments in Chromatic

winkerVSbecks:
@domyen updated with changes you requested.

domyen (approved):

kylesuss:

PR #214, opened 2020-10-17
Fix Loading AvatarList Not Showing in Stories

ahmadatallah:
**Description**

A tiny PR for showing the loading avatar-list example in stories as it was not showing and instead, it displays empty avatar-list.

**Checklist**

- [x] Add users and userCount to the loading avatar story. 
- [x] Fix prettier warning for always add parentheses around arrow function (fixed automatically when I run prettier command on the file from my text editor)

### DIFF
--- a/src/components/AvatarList.stories.js
+++ b/src/components/AvatarList.stories.js
@@ -34,7 +34,7 @@ export default {
   excludeStories: ['users'],
 };
 
-export const Basic = args => <AvatarList {...args} />;
+export const Basic = (args) => <AvatarList {...args} />;
 Basic.args = { users };
 
 export const Short = Basic.bind();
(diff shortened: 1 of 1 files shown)


src/components/AvatarList.stories.js
```diff
@@ -50,7 +50,7 @@ export const SmallSize = Basic.bind();
 SmallSize.args = { users, userCount: 100, size: 'small' };
 
 export const Loading = Basic.bind();
-Loading.args = { users: [], isLoading: true };
+Loading.args = { users, userCount: 100, isLoading: true };
```
kylesuss, line 53:
```suggestion
Loading.args = { users: undefined, userCount: undefined, isLoading: true };
```

Just took a look at this and I think the `users` and `userCount` are meant to be left undefined for the loading state since in theory we wouldn't know these values during loading.

kylesuss (changes requested):
Thanks for the PR! I left a comment about the story args -- let me know what you think.

src/components/AvatarList.stories.js
ahmadatallah: yes, makes sense. but then, it should be only 
```suggestion
Loading.args = { users: undefined, isLoading: true };
```
as the `userCount` prop is optional!

kylesuss: Nice one. Sounds good!

kylesuss (approved):
💯  Thanks for this!

kylesuss:

PR #215, opened 2020-10-19
Convert Link to TS

winkerVSbecks:


src/components/Link.stories.tsx
```diff
@@ -19,7 +20,7 @@ export default {
   component: Link,
 };
 
-export const Basic = (args) => <Link {...args} />;
+export const Basic = (args: LinkProps) => <Link {...args} />;
```
ndelangen, line 23: please prefer using the FunctionComponent<Props>

```suggestion
export const Basic: FunctionComponent<LinkProps> = (props) => <Link {...props} />;
```

src/components/Link.tsx
```diff
@@ -127,78 +135,44 @@ const LinkButton = styled.button`
...
-  ) => {
-    // Use the UnstyledLink here to avoid duplicating styles and creating
-    // specificity conflicts by first rendering the StyledLink higher up the chain
-    // and then re-rendering it through the 'as' prop.
+/**
+ * Links can contains text and/or icons. Be careful using only icons, you must provide a text alternative via aria-label for accessibility.
+ */
+export type LinkProps = React.ComponentPropsWithoutRef<'a'> &
+  React.ComponentPropsWithoutRef<'button'> &
+  StyledLinkProps & {
```
ndelangen, line 143: You should be able to use `ComponentProps<StyledLink>`

ndelangen (changes requested):

kylesuss:
I defer to @ndelangen on this as he is much more experienced w/ TS than I.

ndelangen:
LMK if you want to pair on this @winkerVSbecks ?

src/components/Link.tsx
winkerVSbecks: I couldn't get that working. I get an error: `'StyledLink' refers to a value, but is being used as a type here. Did you mean 'typeof StyledLink'`

winkerVSbecks:
@ndelangen updated with your suggestions. LMK if this looks good, otherwise would love to pair.

src/components/Link.stories.tsx
```diff
@@ -3,7 +3,8 @@ import styled from 'styled-components';
 import { action } from '@storybook/addon-actions';
 
 import { Icon } from './Icon';
-import { Link } from './Link';
+import { Link, LinkProps } from './Link';
```
ndelangen, line 6: if possible use `ComponentProps<typeof Link>` over exporting `LinkProps`

src/components/Link.tsx
```diff
@@ -1,12 +1,12 @@
+/* eslint-disable react/prop-types */
```
ndelangen, line 1: We can tell eslint that this rule doesn't need to be applied at all on .tsx files

src/components/Link.tsx
```diff
@@ -21,7 +21,15 @@ const LinkInner = styled.span`
...
 `;
 
-const StyledLink = styled.a`
+type StyledLinkProps = {
+  containsIcon?: boolean;
+  secondary?: boolean;
+  tertiary?: boolean;
+  nochrome?: boolean;
+  inverse?: boolean;
+};
```
ndelangen, line 30:
```suggestion
interface StyledLinkProps {
  containsIcon?: boolean;
  secondary?: boolean;
  tertiary?: boolean;
  nochrome?: boolean;
  inverse?: boolean;
};
```

ndelangen (approved):

src/components/Link.tsx
ndelangen: Looks great, some optional improvements, if agree with them 🎉  👍

winkerVSbecks:
@ndelangen one more update. Added the `LinkComponentPicker` back in. Removing it caused unnecessary props to be padded into the `LinkWrapper`.

ndelangen (approved):

ndelangen:
good to merge!

kylesuss:

PR #216, opened 2020-10-23
Input styling changes, Form component addition

kylesuss:
The `Input` element has new styling, per the designs here:

https://projects.invisionapp.com/d/main#/console/20597416/432876968/preview

In addition to that, I have added a "lightweight"`Form` component that does not make any assumptions about how you manage form state (values, onChange, etc.), though it does encode some UX patterns for how we handle error messages & validate form fields. The intent was to make this extremely flexible, but create a set of behaviors that we will want on form across our properties based on the design language. I added some example for how to manage state in the stories, but they are mostly meant for documentation/example purposes.

The API is backwards compatible, so I am setting this to a `minor` release.

PS this will be the first design system component I added w/ TypeScript so let me know what I can do to improve that 👍


src/components/tooltip/Tooltip.js
```diff
@@ -80,14 +87,14 @@ export function Tooltip({
 }) {
   return (
     <TooltipWrapper hasChrome={hasChrome} data-placement={placement} ref={tooltipRef} {...props}>
-      {hasChrome && <Arrow data-placement={placement} ref={arrowRef} {...arrowProps} />}
+      <Arrow isVisible={hasChrome} data-placement={placement} ref={arrowRef} {...arrowProps} />
```
winkerVSbecks, line 90: Thank you for adding this one!

src/components/Form.tsx
```diff
@@ -0,0 +1,91 @@
+import React, { useEffect, useState } from 'react';
+
+interface Field {
+  id: string;
+  validationError: (value: string) => string | void;
+  Component: (props: any) => JSX.Element;
```
winkerVSbecks, line 6: I believe `ComponentType` might be more appropriate here. @ndelangen wdyt?

```ts
interface Field<T> {
  id: string;
  validationError: (value: string) => string | void;
  Component: React.ComponentType<T>
}
```

winkerVSbecks (commented):
Not sure about that one typing but, other LGTM

src/components/tooltip/Tooltip.js
kylesuss: No prob -- did you run into this issue as well?

winkerVSbecks: yea, while adding the search dropdown on the blog

src/components/Form.tsx
ndelangen: yeah!

ndelangen (approved):

kylesuss:
@ghengeveld I reworked this w/ the pattern we discussed on the call today. I think it ends up being more flexible, as you say. LMK what you think.

src/components/FormErrorState.tsx
```diff
@@ -56,7 +57,7 @@ export const PureFormErrorState = ({
   return children({ getFormErrorFieldProps });
 };
 
-export const FormErrorState = (props: FormErrorStateProps) => {
+export const FormErrorState = ({ didAttemptSubmission = false, ...rest }: FormErrorStateProps) => {
```
ndelangen, line 60:
```suggestion
export const FormErrorState: FunctionComponent<FormErrorStateProps> = ({ didAttemptSubmission = false, ...rest }) => {
```

kylesuss: Nice one. Thanks.

ndelangen (approved):

ghengeveld (approved):

ndelangen:
👏

kylesuss:

PR #221, opened 2020-11-17
Port over Cardinal component from Chromatic

winkerVSbecks:
@domyen ported over as is. The only change is left aligning both small and large cardinal.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,9 @@
     ]
   },
   "dependencies": {
+    "@types/pluralize": "^0.0.29",
     "copy-to-clipboard": "^3.3.1",
+    "pluralize": "^8.0.0",
     "polished": "^3.6.4",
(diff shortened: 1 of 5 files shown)


kylesuss (approved):
👍

domyen:
If we want to preserve center alignment for Large cardinals on the BuildScreen and PullRequestScreen, would we override in the Chromatic project or should that be an option here?

winkerVSbecks:
@domyen I would add a prop to control that.

domyen:
lgtm! The naming structure changed so you might want to pull `master` and remove `Design System` from the component title.

domyen (approved):

winkerVSbecks:
:rocket: PR was released in [`v5.3.0`](https://github.com/storybookjs/design-system/releases/tag/v5.3.0) :rocket:

PR #222, opened 2020-11-18
Clipboard components

winkerVSbecks:
* Converted Clipboard to TS
* Ported over ClipboardIcon, ClipboardCode and ClipboardInput

PR #240, opened 2020-12-17
Fix clipboard logic to reset state

kylesuss:
Small bug introduced in the refactor at #222 . The timeout never gets called.

### DIFF
--- a/src/components/clipboard/Clipboard.tsx
+++ b/src/components/clipboard/Clipboard.tsx
@@ -39,7 +39,7 @@ export const Clipboard = ({
     if (copied && timeoutId) {
       clearTimeout(timeoutId);
     }
-    if (copied && resetTimeout && timeoutId) {
+    if (copied && resetTimeout) {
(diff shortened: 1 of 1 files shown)


src/components/clipboard/Clipboard.tsx
```diff
@@ -39,7 +39,7 @@ export const Clipboard = ({
     if (copied && timeoutId) {
       clearTimeout(timeoutId);
     }
-    if (copied && resetTimeout && timeoutId) {
+    if (copied && resetTimeout) {
       timeoutId = setTimeout(() => setCopied(false), resetTimeout);
     }
```
kylesuss, line 44: 🐔  & 🥚  problem. The `timeoutId` can't exist without first hitting this code block so we will never end up setting the timeout.

winkerVSbecks (approved):

kylesuss:
:rocket: PR was released in [`v5.4.5`](https://github.com/storybookjs/design-system/releases/tag/v5.4.5) :rocket:

PR #241, opened 2020-12-29
Use focus-visible to hide outlines when using mouse 

domyen:
Hide :focus outlines when using a mouse and remove a bunch of :focus button styles.

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -69,14 +69,6 @@ export const StyledButton = styled.button`
       &:active {
         transform: translate3d(0, 0, 0);
       }
-
-      &:focus {
-        box-shadow: ${rgba(color.primary, 0.4)} 0 1px 9px 2px;
-      }
-
-      &:focus:hover {
-        box-shadow: ${rgba(color.primary, 0.2)} 0 8px 18px 0px;
-      }
     `}
 
   ${Text} {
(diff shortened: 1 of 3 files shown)


src/components/Button.js
```diff
@@ -69,14 +69,6 @@ export const StyledButton = styled.button`
...
         transform: translate3d(0, 0, 0);
       }
-
-      &:focus {
-        box-shadow: ${rgba(color.primary, 0.4)} 0 1px 9px 2px;
-      }
-
-      &:focus:hover {
-        box-shadow: ${rgba(color.primary, 0.2)} 0 8px 18px 0px;
-      }
```
kylesuss, line 79: Doesn't look like you added these styles back anywhere. Do I have that right? Was that the intent?

domyen: That's right. I removed them all because we now have a better way to detect whether something if focused via mouse and via keyboard. If it's via mouse, there's no focus UI. If it's via keyboard, we should use the system default focus UI.

kylesuss: I see. Cool. Makes sense.

src/components/shared/global.ts
```diff
@@ -122,4 +121,9 @@ export const GlobalStyle = createGlobalStyle`
   body {
     ${bodyStyles}
   }
+  // prevent mouse-clicks from focusing elements
+  // this removes the ugly blue outline
+  * :focus:not(:focus-visible) {
```
kylesuss, line 126: Should there be a space in between `*` and `:focus`?

kylesuss (approved):
Seems good. Just want to validate that the space I called out is not causing any unexpected behavior.

src/components/shared/global.ts
domyen: Removed the `* ` because it's not needed after all.

kylesuss:
Merge when ready 👍

kylesuss:
:rocket: PR was released in [`v5.4.6`](https://github.com/storybookjs/design-system/releases/tag/v5.4.6) :rocket:

PR #242, opened 2020-12-29
Improve tooltip stories

domyen:
The tooltip stories should use buttons not divs.

### DIFF
--- a/src/components/tooltip/TooltipLinkList.stories.js
+++ b/src/components/tooltip/TooltipLinkList.stories.js
@@ -18,7 +18,7 @@ export default {
     (storyFn) => (
       <div style={{ height: '300px' }}>
         <WithTooltip placement="top" trigger="click" startOpen tooltip={storyFn()}>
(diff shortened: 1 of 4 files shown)


kylesuss:
There is some logic to this, I believe. A little while back, Jimmy updated `WithTooltip` to render a button already for accessibility reasons. You can see that play out here:

https://github.com/storybookjs/design-system/blob/1957d198f335706b02c5af82ff5d9ed5a20c9d5a/src/components/tooltip/WithTooltip.js#L10

https://github.com/storybookjs/design-system/blob/1957d198f335706b02c5af82ff5d9ed5a20c9d5a/src/components/tooltip/WithTooltip.js#L43

Unless you explicitly pass in a "tagName" to override the default rendering of a `button`, the component already does this for you. With that in mind, it doesn't really make sense to me to render another button as a child of that. Thoughts?

domyen:
That's helpful context! Closing now.

PR #245, opened 2021-01-05
Update tertiary Select variant to match new designs

kylesuss:
As discussed, here are the changes for the tertiary Select variant.


kylesuss:
:rocket: PR was released in [`v5.4.6`](https://github.com/storybookjs/design-system/releases/tag/v5.4.6) :rocket:

PR #246, opened 2021-01-12
Add function support for errors on textarea, startFocused behavior

kylesuss:
This PR adds some logic that was already ported over to inputs. We now support an error function being passed in as well as the old approach of an error string. On top of that, we can have the textarea `startFocused`. Doing this aligns the textarea with the behavior that was setup for form validation recently.

### DIFF
--- a/src/components/Textarea.js
+++ b/src/components/Textarea.js
@@ -52,7 +52,7 @@ const LabelWrapper = styled.div`
 const Subtext = styled.div``;
 
 const TextareaText = styled.textarea`
-  ::placeholder: ${color.medium};
+  ::placeholder: ${color.mediumdark};
   appearance: none;
   border: none;
   box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


src/components/Textarea.js
```diff
@@ -52,7 +52,9 @@ const LabelWrapper = styled.div`
 const Subtext = styled.div``;
 
 const TextareaText = styled.textarea`
-  ::placeholder: ${color.medium};
+  ::placeholder {
+    ${color.mediumdark};
```
domyen, line 56:
```suggestion
    color: ${color.mediumdark};
```

kylesuss: Thought I changed this 🤔 . Anyhow fixed.

https://github.com/storybookjs/design-system/pull/246/commits/b419b6140806ae494662d52a713522d6819f5f53

kylesuss:
:rocket: PR was released in [`v5.4.7`](https://github.com/storybookjs/design-system/releases/tag/v5.4.7) :rocket:

PR #249, opened 2021-02-22
Add organization type to avatar

kylesuss:
This new avatar type is used when showing avatars for organizations. In the loading state, there is an icon and the "user" icon doesn't make sense in this context, hence the change. There is also a slightly different border radius on this type.


domyen (approved):

domyen (approved):

domyen:
Left a request in Chromatic

kylesuss:
:rocket: PR was released in [`v5.4.8`](https://github.com/storybookjs/design-system/releases/tag/v5.4.8) :rocket:

PR #250, opened 2021-02-23
Allow customization of form error state placement

kylesuss:
Fixes: 
Screen Shot 2021-02-23 at 3 36 59 PM

Unfortunately, there isn't a great way to customize the positioning based on browser window & it is a bit of a pain to setup event listeners for this sort of thing. With that in mind, I chose the `auto-end` positioning (see: https://popper.js.org/docs/v1/#popperplacements--codeenumcode) which typically means that the tooltip shows up underneath the input and at the very right side of it. I think this is better than just "auto" which places it on the bottom & centered, but maybe it is just a personal preference.

@domyen thoughts on this? We seem to be slightly limited by what the lib will let us do.

### DIFF
--- a/src/components/Input.js
+++ b/src/components/Input.js
@@ -5,7 +5,7 @@ import { color, typography, spacing } from './shared/styles';
 import { jiggle } from './shared/animation';
 import { Icon } from './Icon';
 import { Link } from './Link';
-import WithTooltip from './tooltip/WithTooltip';
+import WithTooltip, { validPlacements as validTooltipPlacements } from './tooltip/WithTooltip';
 import { TooltipMessage } from './tooltip/TooltipMessage';
 
 // prettier-ignore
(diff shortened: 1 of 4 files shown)


kylesuss:
@domyen Hold that thought -- the new lib version might have a new setting that we need but I have to do some work to upgrade.

kylesuss:
@domyen ok this has been updated w/ the approach that I showed you earlier. The design system now allows `Input`s to customize the `errorTooltipPlacement`. In consumers of the design system, logic can be setup to change the placement based on screen size, etc.

src/components/Input.js
```diff
@@ -357,6 +358,7 @@ PureInput.propTypes = {
   id: PropTypes.string.isRequired,
   value: PropTypes.string,
   appearance: PropTypes.oneOf(['default', 'secondary', 'tertiary', 'pill', 'code']),
+  errorTooltipPlacement: PropTypes.string,
```
domyen, line 361: There are a fixed number of placements right? Should we encode them here instead of allowing for any ol' string?

That way they also get auto-documented in Storybook.

kylesuss: There's way more than you might think. You can add variations like `right-end` and `right-start` in addition to the main top, left, right, auto, etc. so it feels like overkill to define them all. WDYT?

https://popper.js.org/docs/v1/#popperplacements--codeenumcode

kylesuss: @domyen Nvm it wasn't too bad I guess:

https://github.com/storybookjs/design-system/pull/250/commits/594a6af81c239f796fa03f68ed5e10bf7743203b

kylesuss:
:rocket: PR was released in [`v5.4.9`](https://github.com/storybookjs/design-system/releases/tag/v5.4.9) :rocket:

PR #255, opened 2021-03-31
Update design system

domyen:
**Changelog**
- Input: remove `secondary` state and use `default` state instead
- Input: update `code` state based on the styling we use in Chromatic
- Select: remove `secondary` and use `default` state instead 
- Textarea: remove `code` state
- Textarea: remove `secondary` state and use `default` state instead
- Add and update comment icons
image

**Note**
This requires downstream changes in Chromatic and LearnStorybook to use the new APIs


src/components/Radio.stories.js
```diff
@@ -10,8 +10,8 @@ export default {
   component: Radio,
 };
 
-export const Basic = (args) => <Radio {...args} />;
-Basic.args = { label: 'label', hideLabel: false };
+export const playground = (args) => <Radio {...args} />;
+playground.args = { label: 'label', hideLabel: false };
```
kylesuss, line 14: I feel like we are converging on this pattern of calling these `Template` in the Chromatic repo & also in the SB docs. I haven't seen this "playground" pattern before. Where'd it come from?

domyen: These are the stories that used to be called "Basic". They're the "representative" story that you can manipulate with Controls.

kylesuss: Also the non-capitalization and export (`export playground` vs `export Playground`) goes against this other piece of the docs & pattern we have been following in Chromatic where we do this:

```js
export default {
  includeStories: /^[A-Z]/
}
```

so that all of the capitalized exports are treated as stories.

kylesuss: > These are the stories that used to be called "Basic". They're the "representative" story that you can manipulate with Controls.

That's right. Kinda forgot we had that. Is that a thing we recommend doing with all components these days?

kylesuss (approved):
Seems good. Nice to get rid of the unused stuff. I think it makes sense to establish patterns that we use across codebases so that's why I called out some things here, but I suppose they aren't blocking. We just started in Chromatic focusing a bit more on emulating what the SB docs recommend which probably makes a lot of sense here since its all open source.

src/components/Radio.stories.js
domyen: Useful for design system components, not sure about creating a playground for more complex components

domyen:
@kylesuss can you review this again? If everything is OK please do a minor release 🙏

kylesuss (approved):
Looks good! Appreciate you working through the template/playground stuff w/ me.

kylesuss:
:rocket: PR was released in [`v5.5.0`](https://github.com/storybookjs/design-system/releases/tag/v5.5.0) :rocket:

PR #280, opened 2021-07-30
Support onVisibilityChange in WithTooltip

kylesuss:
React popper supports this handler, so we should as well in case we need to know when the tooltip closes.


coderkevin (approved):
Looks good!

PR #305, opened 2021-09-30
emotion convertion

ndelangen:
@Andarist I made this PR to check if chromatic UI review will report 0 changes as well

<details>
  <summary>📦 Published PR as canary version: <code>6.4.2-canary.305.ab1cc56.0</code></summary>
  <br />
  
  ```bash
  npm install @storybook/design-system@6.4.2-canary.305.ab1cc56.0
  # or 
  yarn add @storybook/design-system@6.4.2-canary.305.ab1cc56.0
  ```
</details>


patches/react-element-to-jsx-string+14.3.2.patch
```diff
@@ -0,0 +1,70 @@
...
+ 
+@@ -443,7 +447,7 @@ var formatComplexDataStructure = (function (value, inline, lvl, options) {
+       }
+ 
+       return originalResult;
+-    }
++    },
+   });
+ 
+   if (inline) {
```
ndelangen, line 70: This is also in a PR, right? can we keep track of that so we can remove this patch some day?

src/components/AvatarList.tsx
```diff
@@ -25,6 +25,19 @@ const UserEllipses = styled.li`
...
+  }
+  &:nth-child(1) {
+    z-index: 3;
+  }
+  &:nth-child(2) {
+    z-index: 2;
+  }
+  &:nth-child(3) {
+    z-index: 1;
+  }
```
ndelangen, line 40: I think this change should be reverted maybe?

I think i did this trying to get stuff to work, but with your changes, the previous code should just work, right?

src/components/shared/global.tsx
```diff
@@ -121,7 +122,7 @@ export const bodyStyles = css`
 export const fontUrl =
   'https://fonts.googleapis.com/css?family=Nunito+Sans:400,700,800,900&display=swap';
 
-export const GlobalStyle = createGlobalStyle`
+const newLocal = css`
```
ndelangen, line 125: Should get a proper name, I guess

src/components/tooltip/ListItem.tsx
```diff
@@ -98,56 +101,71 @@ const linkStyles = css<{
...
+      activeColor,
+      ...linkWrapperRest
+    }: ComponentProps<LinkWrapperType> & LinkStyleProps) => <LinkWrapper {...linkWrapperRest} />
+  )`
+    ${linkStyles}
+  `
+);
+
+type StyledLinkWrapperProps = ComponentProps<ReturnType<typeof buildStyledLinkWrapper>>;
+
```
ndelangen, line 167: This looks complicated, can we add a code-comment explaining what this is doing and why it's required?

ndelangen:
Overall, looks good! I'm confident we can proceed.

src/components/AvatarList.tsx
```diff
@@ -25,6 +25,19 @@ const UserEllipses = styled.li`
 
 const User = styled.li`
   display: inline-flex;
+
+  &:not(:first-child) {
```
kylegach, line 29: Will these components never be SSR'd? I thought emotion didn't like `:first-child`, `:nth-child`, etc. At the very least, I'm pretty sure this will log warnings in the console.

src/components/AvatarList.tsx
Andarist: gonna take a look at reverting this change 👍

src/components/AvatarList.tsx
Andarist: yes, this might produce warnings in the console - since you don't expect this design system to be consumed publicly I think it's acceptable to just wrap your application in the CacheProvider with a "compat" cache. This allows you to turn off those warnings for the whole app - if you are going to implement SSR anyway you probably won't use streaming anyway so you can use `@emotion/server` to extract things and construct the final HTML in a way that those warnings won't be relevant for you (you won't have problems by using that kind of selectors).

Keep in mind that React is working on the new API now (for React 18) that might allow us to circumvent this issue altogether. My hope is that it will allow us to stream without creating the `:first-child`~ problem. You can check out this PR to get some more information: https://github.com/facebook/react/pull/21913

patches/react-element-to-jsx-string+14.3.2.patch
Andarist: The PRs in question are mentioned above those patches - should I list them anywhere else? One of those has already landed (so I will be able to remove this part of the patch soon) and the other one is blocked because I can't create a test case for it. After I've fixed the issue it went away and I'm unsure what exactly has triggered it. Gonna try removing this part of the patch and observe if it comes back - so maybe I will run into this again.

src/components/tooltip/ListItem.tsx
Andarist: Sure - gonna add this in a comment later. 

The gist is that you are creating a final component type dynamically in the render function here:
https://github.com/storybookjs/design-system/blob/e85d0017be39c2d62c393d83617d8948d496e0af/src/components/tooltip/ListItem.tsx#L189
which means that the created component will lose its state with each rerender. React has to unmount and remount this because it can't recognize that it's the same component type (its identity has changed after all). `weakMemoize` just ensures that you always return the same component type for the given input component type. This package is already a part of your bundle because it's used internally by Emotion itself (but I should probably downgrade it to the version used by Emotion 10) - so this doesn't add much to your bundlesize.

This change was unrelated to the Emotion migration but is a small fix that might spare you some potential debugging time in the future.

src/components/AvatarList.tsx
kylegach: Sounds good. Thanks for the info, Mateusz.

Fwiw, while it may not be promoted as such, the SB design system _is_ consumed publicly. I know because I was doing so, in the project I was working on before I joined the team here. It's a great way to make things feel Storybook-native.

Andarist: If this is also used publicly then I would recommend making this a major change either way because SSRing strategies might change, some subtle differences in browsers might also be present.

As to the whole warning thing - it depends on how do you want to handle that and what kind of guarantees/recommendations do you want to give to your consumers. For instance "setting up" doc that would describe the CacheProvider setup and the SSR strategy could be OK **if** you think that this is enough. If you want to avoid that (but you can't easily avoid the latter - SSR strategies) then you can use the silencing comment, like you are already doing in the main Storybook repository.

I understand that this issue is somewhat annoying for people relying on those kinds of selectors - but hopefully it won't be a problem in the future (as mentioned in the previous comment).

README.md
```diff
@@ -66,7 +66,7 @@ const { GlobalStyle } = global;
 Useful when you want build upon the shared global styling.
 
 ```javascript
-import { createGlobalStyle } from 'styled-components';
+import { createGlobalStyle } from '@storybook/theming';
```
Andarist, line 69: note for myself: this should be fixed

ndelangen:
@kylegach @winkerVSbecks What's the process of releasing this as a new major version?

Is it easy for either of you to do?

Will merging this cause it the be released as a new major automatically?

kylegach:
> @kylegach @winkerVSbecks What's the process of releasing this as a new major version?
> 
> Is it easy for either of you to do?
> 
> Will merging this cause it the be released as a new major automatically?

I believe it's based on labels. I just added a "Major" label, so it _should_ do the right thing when merged. Someone should confirm my understanding on that, though.

PR #315, opened 2021-11-08
Remove props from Icon

smithambera:
Fixes https://linear.app/chromaui/issue/CH-876/remove-color-prop-from-ds-icon

## Description

Feedback from @domyen:

> Our current ergonomics are to define styles outside of the component markup (i.e., in styled and not inline). It seems like we’re moving to that even more with the *.style.ts pattern I saw in one of your PRs. I don’t think the Icon should have a color prop because it’s a departure from these ergonomics.

As such, I removed the `color` prop. I also removed the `block` prop for the same reason - it didn't seem to be used anywhere anyway.

<details>
  <summary>📦 Published PR as canary version: <code>6.4.2-canary.315.470be8c.0</code></summary>
  <br />
  
  ```bash
  npm install @storybook/design-system@6.4.2-canary.315.470be8c.0
  # or 
  yarn add @storybook/design-system@6.4.2-canary.315.470be8c.0
  ```
</details>

### DIFF
--- a/src/components/Icon.tsx
+++ b/src/components/Icon.tsx
@@ -2,17 +2,14 @@ import React, { FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
 import { icons } from './shared/icons';
 
-const Svg = styled('svg', { shouldForwardProp: (prop) => !['block', 'color'].includes(prop) })<
-  Partial<Props>
->`
-  display: ${(props) => (props.block ? 'block' : 'inline-block')};
+const Svg = styled.svg`
   vertical-align: middle;
 
   shape-rendering: inherit;
   transform: translate3d(0, 0, 0);
 
   path {
-    fill: ${(props) => props.color};
+    fill: currentColor;
   }
 `;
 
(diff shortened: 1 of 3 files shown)


src/components/Icon.tsx
```diff
@@ -2,17 +2,14 @@ import React, { FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
 import { icons } from './shared/icons';
 
-const Svg = styled('svg', { shouldForwardProp: (prop) => !['block', 'color'].includes(prop) })<
-  Partial<Props>
->`
-  display: ${(props) => (props.block ? 'block' : 'inline-block')};
+const Svg = styled.svg`
```
domyen, line 5: I think we still want `display: inline-block;` to be set explicitly here because we often use icons side-by-side with `inline` text elements.

domyen (changes requested):
Thanks Amber, this looks good. One blocking comment (no pun intended) 🤗

src/components/Icon.tsx
smithambera: Yap, forgot to add it back! My b.

domyen (approved):

domyen:
lgtm! merging

PR #321, opened 2021-11-17
Pulled in `AuthorList` component from `frontpage`

smithambera:
Fixes https://linear.app/chromaui/issue/CH-710/pull-author-and-authorlist-from-frontpage-into-ds

## Description

Pulled the `AuthorList` component from https://github.com/storybookjs/frontpage into the DS in preparation for future work.

### QA instructions

Confirm the styles (mostly) match [the designs]((https://www.figma.com/file/gYdEmX9zYowUR3K2SOua5J/Component-catalog?node-id=148%3A9975).

<details>
  <summary>📦 Published PR as canary version: <code>7.0.4-canary.321.a50c445.0</code></summary>
  <br />
  
  ```bash
  npm install @storybook/design-system@7.0.4-canary.321.a50c445.0
  # or 
  yarn add @storybook/design-system@7.0.4-canary.321.a50c445.0
  ```
</details>

### DIFF
--- a/src/components/AuthorList.stories.tsx
+++ b/src/components/AuthorList.stories.tsx
@@ -33,6 +33,11 @@ export const authors = [
     name: 'Kyle Suss',
     avatarUrl: 'https://avatars2.githubusercontent.com/u/3035355?s=96&v=4',
   },
+  {
+    id: '7',
+    name: 'Kyler Suspect',
+    avatarUrl: 'https://avatars2.githubusercontent.com/u/3035355?s=96&v=4',
+  },
 ];
 
 export default {
@@ -47,5 +52,8 @@ Basic.args = { authors: authors.slice(0, 5) };
 export const Short = Basic.bind({});
 Short.args = { authors: authors.slice(0, 2) };
 
-export const ShowMore = Basic.bind({});
-ShowMore.args = { authors };
+export const ShowOneMoreAuthor = Basic.bind({});
+ShowOneMoreAuthor.args = { authors: authors.slice(0, 6) };
+
+export const ShowMultipleMoreAuthors = Basic.bind({});
+ShowMultipleMoreAuthors.args = { authors };
--- a/src/components/AuthorList.stories.tsx
+++ b/src/components/AuthorList.stories.tsx
@@ -1,0 +1,51 @@
+import React from 'react';
+import { AuthorList, AuthorListProps } from './AuthorList';
+
+export const authors = [
+  {
+    id: '1',
+    name: 'Dominic Nguyen',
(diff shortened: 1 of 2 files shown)


domyen:
Thanks Amber! 

There are a few cases to refine that weren't covered in the first buildout. We should style and make stories for them while we're working on this component.

1. Show more interactive
2. Show more non-interactive (can happen when we're querying a 3rd party API like NPM)

[Figma file here](https://www.figma.com/file/gYdEmX9zYowUR3K2SOua5J/Component-catalog?node-id=1166%3A20098)

image

smithambera:
Hey @domyen - can you send me that illustration? I didn't see it in `frontpage` or `chromatic`. I did find `avatar.png` in `chromatic`, but it's not quite the same:

Screen Shot 2021-11-20 at 12 34 29 PM

Did you want ^ that illustration ^ when there is only one author hidden and the one from the designs above if there is more than one author hidden? Also, it'd probably be preferable to get them as svgs!

domyen:
Thanks for the callout Amber and amazing idea to have 1 user vs 2 users depending on how many are left to reveal.

These images are in the `<Icon>` component as `user` and `users` respectively.

smithambera:
Hey @domyen - it doesn't look like the current component supports interactivity for the hidden authors. What is your desired behavior there?

Additionally, based on the way the component is presently built, there would be no way to see the non-interactive state as we're expecting this component to get the entire list of authors from the get-go.

domyen:
> it doesn't look like the current component supports interactivity for the hidden authors. What is your desired behavior there?
If we _can_ get the full list of people, the intended behavior is when you click on the "+# more" that the full list of people is rendered in place of the "+# more".

> Additionally, based on the way the component is presently built, there would be no way to see the non-interactive state as we're expecting this component to get the entire list of authors from the get-go.
Some services like NPM do not give us the full list of people afaict. Check out what's happening in the "Made by" section of [this addon page](https://storybook.js.org/addons/@storybook/addon-postcss/).

smithambera:
> Some services like NPM do not give us the full list of people afaict. Check out what's happening in the "Made by" section of this addon page.

The existing `AuthorList` code is expecting all of the authors and only showing 5:

```
const authorsSubset = useMemo(() => authors.slice(0, 5), [authors]);
const more = useMemo(() => authors.length - 5, [authors]);
```

In that link you shared, the component has access to all of the authors:
Screen Shot 2021-12-01 at 2 05 09 PM

domyen:
Doh! Thanks for clearing that up Amber – let's remove the non-interactive state.

PR #331, opened 2022-01-20
Improve Select styling and fix Safari bug

domyen:
- Update `Select` styling to have a blue focus color
- Fix a Safari rendering bug that @zol reported
- Update stories to be more inline with other `Input` stories

OLD
image

NEW
image


domyen:
Self-merging as this is a minor improvement. It was tested in all browsers.

PR #364, opened 2022-06-23
Update every icon (and add a few new ones)

MichaelArestad:
1. Every icon has been redrawn at 14px x 14px.
2. Several new/missing icons have been added.
3. The Icon component has been rewritten to allow multiple paths.

**Bonus:** `Icons.tsx` (formerly `icons.ts`) went from 129,864 bytes to 68,898 bytes even though I added 20 additional icons.

<details>
  <summary>📦 Published PR as canary version: <code>7.4.2-canary.364.5f00c3a.0</code></summary>
  <br />
  
  ```bash
  npm install @storybook/design-system@7.4.2-canary.364.5f00c3a.0
  # or 
  yarn add @storybook/design-system@7.4.2-canary.364.5f00c3a.0
  ```
</details>


src/components/Icon.tsx
```diff
@@ -22,12 +22,12 @@ const Svg = styled.svg`
  */
 export const Icon: FunctionComponent<Props> = ({ icon, ...props }: Props) => {
   return (
-    <Svg viewBox="0 0 1024 1024" width="20px" height="20px" {...props}>
-      <path d={icons[icon]} />
+    <Svg viewBox="0 0 14 14" width="14px" height="14px" {...props}>
+      {Array.isArray(icons[icon]) ? <Fragment>{icons[icon]}</Fragment> : <path d={icons[icon]} />}
```
kylegach, line 26: If you make the [change I proposed to `icons.tsx`](https://github.com/storybookjs/design-system/pull/364/files#r905448974), this can be simplified to:

```jsx
<Fragment>{icons[icon]}</Fragment>
```

_Note that you don't need to import `Fragment` if you use the `<></>` syntax, instead, which is generally preferred._

As a bonus, this also addresses the TypeScript error reported for this file.

src/components/Icon.tsx
```diff
@@ -22,12 +22,12 @@ const Svg = styled.svg`
...
     </Svg>
   );
 };
 
 interface Props {
-  icon: keyof typeof icons;
+  icon?: keyof typeof icons;
```
kylegach, line 32: Why is this optional? It doesn't have a default value, so it would render `<path d={undefined} />` if no `icon` is provided.

src/components/shared/icons.tsx
```diff
@@ -0,0 +1,907 @@
...
+  users: [
+    <path d="M9.21 11.62A10.59 10.59 0 0 1 0 11.07c.06-1.35 2.93-2.58 3.7-2.7.25-.03.26-.68.26-.68s-.72-.69-.87-1.6c-.42 0-.68-.99-.26-1.33 0-.03 0-.08-.02-.14-.07-.63-.3-2.71 2.12-2.71 2.41 0 2.18 2.08 2.11 2.71l-.01.14c.42.34.16 1.32-.26 1.32-.16.92-.87 1.6-.87 1.6s0 .66.25.7c.78.11 3.7 1.36 3.7 2.73v.22l-.64.3Z" />,
+    <path d="M8.81 8.42a9.64 9.64 0 0 0-.74-.4 5.2 5.2 0 0 1 1.7-.76c.17-.02.17-.47.17-.47s-.49-.47-.6-1.1c-.28 0-.46-.68-.17-.91l-.01-.1c-.05-.43-.2-1.86 1.45-1.86 1.66 0 1.5 1.43 1.45 1.86v.1c.28.23.1.9-.18.9-.11.64-.6 1.11-.6 1.11s0 .45.17.47c.54.08 2.55.94 2.55 1.89v.62a10.6 10.6 0 0 1-3.3.56 2.97 2.97 0 0 0-.58-.88c-.37-.41-.85-.76-1.31-1.03Z" />,
+  ],
+  profile: [
+    <path d="M9.1 7.35a5.06 5.06 0 0 1-4.52-.28C4.6 6.4 6.02 5.77 6.4 5.7c.12-.02.12-.35.12-.35s-.35-.34-.43-.81c-.2 0-.33-.5-.12-.67l-.01-.07C5.93 3.48 5.81 2.42 7 2.42S8.07 3.48 8.04 3.8v.07c.2.17.07.67-.13.67-.08.47-.43.81-.43.81s0 .33.12.35c.38.06 1.82.7 1.82 1.4v.1c-.1.06-.2.1-.31.15Zm-5.35 3.9c0-.14.11-.25.25-.25h6a.25.25 0 1 1 0 .5H4a.25.25 0 0 1-.25-.25ZM4 9a.25.25 0 0 0 0 .5h6a.25.25 0 1 0 0-.5H4Z" />,
+    <path
+      fill-rule="evenodd"
```
kylegach, line 17: JSX uses camelCase attributes, so this should be `fillRule`.

src/components/shared/icons.tsx
```diff
@@ -0,0 +1,907 @@
+import React from 'react';
+
+// Icon paths
+export const icons = {
+  user: 'M14 7A7 7 0 1 1 0 7a7 7 0 0 1 14 0ZM2.67 11.15c.7-1 2.6-1.81 3.2-1.9.22-.04.23-.66.23-.66s-.67-.66-.81-1.55c-.4 0-.63-.94-.24-1.27l-.02-.13c-.06-.6-.28-2.6 1.97-2.6s2.03 2 1.97 2.6l-.02.13c.4.33.15 1.27-.24 1.27-.14.89-.8 1.55-.8 1.55s0 .62.22.66c.6.09 2.5.9 3.2 1.9a6 6 0 1 0-8.66 0Z',
+  useralt:
+    'M7.27 13.16a11.39 11.39 0 0 0 5.18-1.23v-.25c0-1.57-3.24-3-4.1-3.13-.27-.05-.28-.79-.28-.79s.8-.78.96-1.83c.47 0 .75-1.12.29-1.52.02-.41.6-3.25-2.32-3.25S4.65 4 4.67 4.41c-.46.4-.17 1.52.29 1.52.17 1.05.96 1.83.96 1.83s0 .74-.27.79c-.86.13-4.04 1.53-4.1 3.08a11.44 11.44 0 0 0 5.72 1.53Z',
+  useradd:
```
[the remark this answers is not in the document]
MichaelArestad: I could definitely do that. I don't have a strong opinion either way. Would you prefer I implement this suggestion?

src/components/shared/icons.tsx
MichaelArestad: Good catch. I'll change all of these.

src/components/Icon.tsx
thafryer: We should remove for a couple reasons. 
1. In terms of our icon usage, there's not really a case where we have an icon that gets used so often that it should be the default.
2. If you do not have the icon prop, it is not explicit which icon is being used which is not helpful when reviewing the code as a dev.

src/components/shared/icons.tsx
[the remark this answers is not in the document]
thafryer: Either solution works, but since you're using JSX, the suggestion Kyle provided will be more readable and have more predictable behavior.

src/components/Icon.tsx
```diff
@@ -1,4 +1,4 @@
-import React, { FunctionComponent } from 'react';
+import React, { Fragment, FunctionComponent } from 'react';
 import { styled } from '@storybook/theming';
 import { icons } from './shared/icons';
```
winkerVSbecks, line 4: How do you feel about adding `color: ${(props) => props.color};` to the SVG component? The new nav uses coloured variants of the icons. This way we can set the color via a prop.

src/components/shared/icons.tsx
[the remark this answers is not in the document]
kylegach: Yes, the benefits explained by Jarel, combined with the benefits of more simple code and eliminating a type error in the Icon.tsx file, make it well worth the effort, imo.

src/components/Icon.tsx
kylegach: Yes, please remove.

src/components/Icon.tsx
domyen: I find it a bit strange to pass styles as props because they end up in the markup part of JSX and not the styles.

src/components/Link.tsx
```diff
@@ -10,8 +10,8 @@ const LinkInner = styled.span<{ withArrow: boolean }>`
     props.withArrow &&
     css`
       > svg:last-of-type {
-        height: 0.7em;
-        width: 0.7em;
+        height: 0.65em;
```
domyen, line 13: Good idea

src/components/Icon.tsx
winkerVSbecks: I hear you. However, this is the alternative and it gets pretty verbose.

```tsx
import { Icon } from '@storybook/design-system';
import { color } from './shared/styles';

export const WhyStorybookIcon = styled(Icon)<{ color: string }>`
  color: ${color.primary};
`;

export const UseCasesIcon = styled(Icon)<{ color: string }>`
  color: ${color.gold};
`;

export const CaseStudiesIcon = styled(Icon)<{ color: string }>`
  color: ${color.green};
`;
```

We could do a typed version. In that case you're specifying token values and not raw CSS.

```tsx
import { styled } from '@storybook/theming';
import { Icon } from '@storybook/design-system';
import { color } from './shared/styles';

export const ColoredIcon = styled(Icon)<{ color: keyof typeof color }>`
  color: ${(props) => color[props.color]};
`;
```

src/components/Icon.tsx
ndelangen: Yea, it's not an optional prop. I think we made it option together when we were adding children, which we ended up not doing at all.

ndelangen:
Do you still need my review? Seems y'all have it pretty well covered tbh 😄

MichaelArestad:
@ndelangen No thank you.

winkerVSbecks (approved):

domyen (approved):

domyen:
lgtm

PR #375, opened 2022-07-13
Format icon elements so react treats them as a the children of an svg

JonathanKolnik:
There is a console error `Warning: Each child in a list should have a unique "key" prop.`

This adds a key to each path child within an icon

<details>
  <summary>📦 Published PR as canary version: <code>7.7.7-canary.375.d1c789b.0</code></summary>
  <br />
  
  ```bash
  npm install @storybook/design-system@7.7.7-canary.375.d1c789b.0
  # or 
  yarn add @storybook/design-system@7.7.7-canary.375.d1c789b.0
  ```
</details>


JonathanKolnik:
the test suite isn't as strict here and didn't fail even though the story logged a console warning, maybe we should update that because it's going to always be a problem if you pull the version into an app that does have that stricter requirement and then you end up with needing a patch like this PR.

JonathanKolnik:
@MichaelArestad the linter complains less if we follow this pattern for the icons, is that ok with you?

kylesuss (approved):
Nice. Haha I think you only needed to change the ones w/ 2 children but this should be good too. Hopefully that didn't make it more complicated. Thanks for switching it up.

PR #376, opened 2022-07-15
Link, Button: Rename prop containsIcon to containsIconOnly 

domyen:
**What?**
Link and Button have a prop called `containsIcon`. This prop is intended for situations where these components **only** contain an icon (no children). But the naming is misleading because you'd think you could apply it when there are children present (but that ends up messing with the alignment).

This PR seeks to improve the naming so reusing the Link and Button components is clearer.

@kylesuss can you review and guide me on whether this is a `patch`, `minor`, or `major` release?

### DIFF
--- a/src/components/Button.stories.tsx
+++ b/src/components/Button.stories.tsx
@@ -109,12 +109,12 @@ export const Disabled = () => (
   </>
 );
 
-export const ContainsIcon = () => (
+export const ContainsIconOnly = () => (
   <>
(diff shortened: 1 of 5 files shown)


kylesuss (approved):
Makes sense, I agree it can be confusing. I would venture to say this is a major release since it is a breaking change & requires changes in the external consumers.

MichaelArestad (approved):

PR #406, opened 2023-02-02
Update browser image with Edge instead of IE

MichaelArestad:
#### Before
image

#### After
image

@thafryer Is this something we should hold off on merging until the new capture stack is available?


thafryer:
It's fine to merge here since we are pinning the design system dependency in Chromatic.

thafryer (approved):

